PR #5, opened 2020-10-19
add search filter mapping

justinpjose:
### What

- Add a mapping filter function so that the filter query from the renderer is changed accordingly for `dp-search-query` to be then passed to the `dp-search-query` itself

### How to review

- Check if the code makes sense
- Check the tests to see if they pass
- Can test locally by running web instances, dp-search-query, dp-frontend-search-controller and then on the browser, go to `http://localhost:20000/search?q=housing&filter=bulletin&filter=article`
- The search results will provide results of `housing` filtered to `bulletin` and `article` types
- Remove `&filter=article` from URL to then only give `bulletin` type results (second test)

### Who can review

Anyone - Justin made the changes


handlers/handlers.go
```diff
@@ -95,3 +94,32 @@ func read(w http.ResponseWriter, req *http.Request, rendC RenderClient, searchC
...
+	filterTypes := map[string]string{
+		"bulletin":              "bulletin",
+		"article":               "article,article_download,static_article",
+		"compendia":             "compendium_landing_page,compendium_chapter",
+		"time_series":           "timeseries",
+		"datasets":              "dataset,dataset_landing_page,compendium_data,reference_tables,timeseries_dataset",
+		"user_requested_data":   "static_adhoc",
+		"methodology":           "static_methodology,static_methodology_download,static_qmi",
+		"corporate_information": "static_foi,static_page,static_landing_page,static_article",
+	}
```
gedge, line 110: better to move this var to a package-wide var, than instantiate each call

handlers/handlers.go
```diff
@@ -95,3 +94,32 @@ func read(w http.ResponseWriter, req *http.Request, rendC RenderClient, searchC
...
+	if len(filters) > 0 {
+		var newFilters string
+		for i, fType := range filters {
+			if (i > 0) && (i <= len(filters)) {
+				newFilters = newFilters + ","
+			}
+			newFilters = newFilters + filterTypes[fType]
+		}
+		query.Del("filter")
+		query.Set("content_type", newFilters)
```
gedge, line 122: using an array and then `Join` is often clearer:
```suggestion
    var newFilters []string
        for _, fType := range filters {
            newFilters = append(newFilters, filterTypes[fType])
        }
        query.Del("filter")
        query.Set("content_type", strings.Join(newFilters, ","))
```

handlers/handlers_test.go
```diff
@@ -120,6 +120,30 @@ func TestUnitHandlers(t *testing.T) {
...
+			apiQuery := mapFilterTypes(query)
+			So(apiQuery["content_type"], ShouldResemble, []string{"article,article_download,static_article,compendium_landing_page,compendium_chapter"})
+		})
+
+		Convey("successfully map no filters given", func() {
+			req := httptest.NewRequest("GET", "/search?q=housing", nil)
+			query := req.URL.Query()
+			apiQuery := mapFilterTypes(query)
+			So(apiQuery["content_type"], ShouldBeNil)
+		})
```
gedge, line 144: see earlier comment
```suggestion
		})
                
                Convey("successfully map bad filters given", func() {
                        req := httptest.NewRequest("GET", "/search?q=housing&filter=INVALID", nil)
                        query := req.URL.Query()
                        apiQuery := mapFilterTypes(query)
                        So(apiQuery["content_type"], ShouldBeNil) // ???
                })
```

handlers/handlers.go
gedge: also, you should cater for the possibility that `filterTypes[fType]` does not exist (return a 400?)

gedge (changes requested):

handlers/handlers_test.go
```diff
@@ -120,6 +120,42 @@ func TestUnitHandlers(t *testing.T) {
...
+
+		Convey("successfully map no filters given", func() {
+			req := httptest.NewRequest("GET", "/search?q=housing", nil)
+			query := req.URL.Query()
+			apiQuery, err := mapFilterTypes(ctx, query)
+			So(apiQuery["content_type"], ShouldBeNil)
+			So(err, ShouldBeNil)
+		})
+
+		// Convey("successfully map bad filters given", func() {
```
gedge, line 150: I think you meant to uncomment this?

justinpjose: Yeap

gedge (approved):

PR #13, opened 2021-02-12
Changes clinging remains of search query to search api

cadmiumcat:
### What

Changes search query to api - the ones we missed in the first round

### How to review

Check that the changes make sense
And that there aren't any liguering references to search query

### Who can review

Anyone


justinpjose (approved):
Looks good to me

PR #15, opened 2021-02-12
Release 0.3.0

cadmiumcat:
- Rename dp-search-query to dp-search-api
- add data.go for search types + add auditing + implement getFilterSortText
- add mapping to count of filter types from search results with tests
- add mapping to count of filter types from search results with tests
- add search filter mapping


nshumoogum (approved):
We probably will need to check whether the new search services are performant enough (I've seen a lot of for loops); they should meet our SLAs

PR #16, opened 2021-02-14
map filter + sort + pagination to search

justinpjose:
### What
- Update sort on the search and add sort data into the controller
- Update filter on the search and add filter data into the controller
- Add pagination control and mapping
- LocaliseKeys are passed to the renderer now rather than values to allow the renderer to then change the values to the preferred language loaded
- Options for limits, sort, and filter are added to the controller instead of the renderer so that the controller controls the data while the renderer just uses that data.
- Validation has been added to limit errors 
- All data for search has a source of truth which is the `data` package. This makes it easier to find search data for any check or update and hence makes it more maintainable.

### How to review
- Check if the tests pass
- Can test locally by running `web` instances, `dp-search-api`, `dp-frontend-search-controller` and then on the browser, go to `http://localhost:20000/search?q=housing`
- The sort, filter and pagination should be functioning correctly - try different use cases
- for `dp-frontend-renderer`, you need to run the branch `develop` to get the latest update on the search page
### Who can review
Anyone


data/filter.go
```diff
@@ -91,3 +100,53 @@ var CorporateInformation = ContentType{
 	Type:            "corporate_information",
 	SubTypes:        []string{"static_foi", "static_page", "static_landing_page", "static_article"},
 }
+
+var errFilterType = errors.New("invalid filter type given")
+
+func setCountZero(categories []Category) []Category {
+	for i, category := range categories {
+		categories[i].Count = 0
```
daiLlew, line 108: what if `category` is nil? This will panic.

data/filter.go
```diff
@@ -91,3 +100,53 @@ var CorporateInformation = ContentType{
...
+		for j := range category.ContentTypes {
+			categories[i].ContentTypes[j].Count = 0
```
daiLlew, line 110: Same here what if `.ContentTypes[j]` is nil?

data/query.go
```diff
@@ -0,0 +1,97 @@
...
+func GetLimitOptions() []int {
+	return []int{10, 25, 50}
+}
+
+// UpdateQueryWithOffset - removes page key and adds offset key to query to be then passed to dp-search-query
+func updateQueryWithOffset(ctx context.Context, query url.Values) url.Values {
+
+	page, err := strconv.Atoi(query.Get("page"))
+	if err != nil {
+		log.Event(ctx, "unable to convert search page to int - set to default "+strconv.Itoa(DefaultPage), log.INFO)
```
daiLlew, line 30: This works but the Go Idiomatic way to do this would be to use the `fmt.Sprintf()` function:
```
log.Event(ctx, fmt.Sprintf("unable to convert search page to int - set to default %d", DefaultPage), log.INFO)
```
It's a bit easier to read (IMO).

data/query.go
```diff
@@ -0,0 +1,97 @@
...
+		page = DefaultPage
+	}
+	limit, err := strconv.Atoi(query.Get("limit"))
+	if err != nil {
+		log.Event(ctx, "unable to convert search limit to int - set to default "+strconv.Itoa(DefaultLimit), log.INFO)
```
daiLlew, line 35: Same here.

data/query.go
```diff
@@ -0,0 +1,97 @@
...
+		limit = DefaultLimit
+	}
+	offset := strconv.Itoa(((page - 1) * limit))
```
daiLlew, line 38: Consider what will happen if `page` & `limit` are negative values. Remember this is taking values provided by a user so we need to account for mistakes and potentially malicious requests.

data/query.go
```diff
@@ -0,0 +1,97 @@
...
+func SetDefaultQueries(ctx context.Context, url *url.URL) *url.URL {
+	var found bool
+	query := url.Query()
+	pageQuery := query.Get("page")
+	if pageQuery == "" {
+		query.Set("page", strconv.Itoa(DefaultPage))
+	} else {
+		page, err := strconv.Atoi(pageQuery)
+		if err != nil {
+			log.Event(ctx, "unable to convert search page to int - default to page "+strconv.Itoa(DefaultPage), log.INFO)
```
daiLlew, line 55: Use sprintf.

data/query.go
```diff
@@ -0,0 +1,97 @@
...
+			query.Set("page", strconv.Itoa(DefaultPage))
+		} else {
+			if page < 1 {
+				log.Event(ctx, "page number is less than 1 - default to page "+strconv.Itoa(DefaultPage), log.INFO)
```
daiLlew, line 59: Same here.

**Observation:** You are using `strconv.Itoa(DefaultPage)` a lot, to me this suggests that it might be beneficial to having a string var of this value
```
var defaultPageStr = "10"
```

data/query.go
nshumoogum: I was wondering if we should even be setting a default value the input is incorrect? Is this the current standard for our controllers?

data/query.go
nshumoogum: Same here too ;)

data/query.go
nshumoogum: Also unnecessary wrapping of brackets

data/filter.go
nshumoogum: categories can't be nil, unless it was updated to be a pointer to []Category, like so:

`func setCountZero(categories []*****Category) []Category {`

data/filter.go
nshumoogum: Same as above

data/query.go
justinpjose: I couldn't find documentation on standards for the controller. But on the current ONS search page, if the user changes the `size` parameter (which is `limit` in our case) to a value less than 10, it will automatically set the limit to 10. This change is achieved by the user changing the `size` value from the URL itself. Otherwise, the user is only limited to changing the value through the dropdown select on the bottom of the search page itself. 

When it comes to the `page` numbers, you will get a Page not Found (404) if the user asks for a page outside the available pages. This change is also similar to the previous

data/query.go
justinpjose: The negative values are already considered in the function `SetDefaultQueries`

data/query.go
justinpjose: Agreed

data/query.go
justinpjose: I have added a string const for those values and hence it is more readable

data/query.go
```diff
@@ -78,6 +79,12 @@ func SetDefaultQueries(ctx context.Context, url *url.URL) (*url.URL, *Pagination
 		query.Set("limit", DefaultLimitStr)
 		limit = DefaultLimit
 	} else {
+		// TO MAKE NEAT
+		if limit > cfg.DefaultMaximumLimit {
+			limit = cfg.DefaultMaximumLimit
+		}
+		if ((limit*currentPage - 1) + 1) > cfg.DefaultMaximumSearchResults {
```
redhug1: There seems to be a line missing between lines 86 & 87  ?

data/query.go
```diff
@@ -78,6 +79,12 @@ func SetDefaultQueries(ctx context.Context, url *url.URL) (*url.URL, *Pagination
 		query.Set("limit", DefaultLimitStr)
 		limit = DefaultLimit
 	} else {
+		// TO MAKE NEAT
+		if limit > cfg.DefaultMaximumLimit {
+			limit = cfg.DefaultMaximumLimit
```
redhug1: Your code coverage / unit tests missed this line.
(they seem to hit nearly everything else, which is great)

data/query.go
```diff
@@ -11,10 +11,14 @@ import (
...
 	// DefaultLimit - default values for limit query
 	DefaultLimit = 10
+	// DefaultLimitStr - default values for limit query in string format
+	DefaultLimitStr = "10"
 	// DefaultSort - default values for sort query
 	DefaultSort = "relevance"
 	// DefaultPage - default values for page query
 	DefaultPage = 1
+	// DefaultPageStr - default values for page query in string format
+	DefaultPageStr = "1"
```
redhug1, line 23: You might consider replacing this 'const' block with the following, which has the advantage of preventing anyone doing maintenance in the future from forgetting to make 'DefaultLimitStr' the equivalent of 'DefaultLimit' and similarly for 'DefaultPageStr & DefaultPage':

const (
	// DefaultLimit - default values for limit query
	DefaultLimit = 10
	// DefaultSort - default values for sort query
	DefaultSort = "relevance"
	// DefaultPage - default values for page query
	DefaultPage = 1
)

// DefaultLimitStr - default values for limit query in string format
var DefaultLimitStr string

// DefaultPageStr - default values for page query in string format
var DefaultPageStr string

func init() {
	DefaultLimitStr = strconv.Itoa(DefaultLimit)
	DefaultPageStr = strconv.Itoa(DefaultPage)
}

handlers/handlers.go
```diff
@@ -128,13 +139,19 @@ func read(w http.ResponseWriter, req *http.Request, rendC RenderClient, searchC
...
+		return
+	}
 	categories, err := getCategoriesTypesCount(ctx, apiQuery, searchC)
 	if err != nil {
 		log.Event(ctx, "mapping count filter types failed", log.Error(err), log.ERROR)
 		setStatusCode(req, w, err)
 		return
 	}
-	err = getSearchPage(w, req, rendC, url, resp, categories)
+	err = getSearchPage(w, req, rendC, url, resp, categories, paginationQuery)
```
redhug1, line 160: I don't think the 'return' on line 158 is needed.

dandwal:
Conflicts with go.mod need to be resolved

config/config.go
```diff
@@ -26,12 +32,18 @@ func Get() (*Config, error) {
 	}
 
 	cfg := &Config{
-		BindAddr:                   ":25000",
-		RendererURL:                "http://localhost:20010",
-		SearchAPIURL:               "http://localhost:23900",
-		GracefulShutdownTimeout:    5 * time.Second,
-		HealthCheckInterval:        30 * time.Second,
-		HealthCheckCriticalTimeout: 90 * time.Second,
+		BindAddr:                    ":25000",
```
jessjenkins, line 35: Suggest you make this localhost so it doesn't cause the mac firewall Allow/Deny popup on startup. This would then need changes to the unit tests as appropriate.
```suggestion
		BindAddr:                    "localhost:25000",
```

config/config.go
```diff
@@ -8,12 +8,18 @@ import (
...
-	GracefulShutdownTimeout    time.Duration `envconfig:"GRACEFUL_SHUTDOWN_TIMEOUT"`
-	HealthCheckInterval        time.Duration `envconfig:"HEALTHCHECK_INTERVAL"`
-	HealthCheckCriticalTimeout time.Duration `envconfig:"HEALTHCHECK_CRITICAL_TIMEOUT"`
+	BindAddr                    string        `envconfig:"BIND_ADDR"`
+	RendererURL                 string        `envconfig:"RENDERER_URL"`
+	SearchAPIURL                string        `envconfig:"SEARCH_API_URL"`
+	GracefulShutdownTimeout     time.Duration `envconfig:"GRACEFUL_SHUTDOWN_TIMEOUT"`
+	HealthCheckInterval         time.Duration `envconfig:"HEALTHCHECK_INTERVAL"`
+	HealthCheckCriticalTimeout  time.Duration `envconfig:"HEALTHCHECK_CRITICAL_TIMEOUT"`
+	DefaultOffset               int           `envconfig:"DEFAULT_OFFSET"`
```
jessjenkins, line 17: Make sure to update the README file with these config items.

go.mod
```diff
@@ -3,17 +3,16 @@ module github.com/ONSdigital/dp-frontend-search-controller
 go 1.15
```
jessjenkins, line 3: This can be upgraded to 1.16 now (and 1.16.1 used for the ci build/unit jobs)

data/query.go
```diff
@@ -0,0 +1,69 @@
...
+
+	reviewSort(ctx, cfg, urlQuery, &validatedQueryParams)
+
+	reviewFilters(ctx, urlQuery, &validatedQueryParams)
+
+	return validatedQueryParams, nil
+}
+
+// GetSearchAPIQuery gets the query that needs to be passed to the search-api to get search results
+func GetSearchAPIQuery(ctx context.Context, cfg *config.Config, validatedQueryParams SearchURLParams) url.Values {
```
dandwal, line 42: config is passed to this function but is never used, was it supposed to be used or should this be removed?

mapper/mapper.go
```diff
@@ -1,203 +1,243 @@
...
+	page.Data.Query = validatedQueryParams.Query
+
+	page.Data.Filter = validatedQueryParams.Filter.Query
+
+	mapSort(page, validatedQueryParams, categories)
+
+	mapPagination(page, validatedQueryParams, categories, respC)
+}
+
+func mapSort(page *model.Page, validatedQueryParams data.SearchURLParams, categories []data.Category) {
```
dandwal, line 32: Categories is passed to this but it is never used, was it supposed to or shall we remove it?

mapper/mapper.go
```diff
@@ -1,203 +1,243 @@
...
+	for _, sort := range data.SortOptions {
+		pageSortOptions = append(pageSortOptions, model.SortOptions{
+			Query:           sort.Query,
+			LocaliseKeyName: sort.LocaliseKeyName,
+		})
 	}
+	page.Data.Sort.Options = pageSortOptions
+}
+
+func mapPagination(page *model.Page, validatedQueryParams data.SearchURLParams, categories []data.Category, respC searchC.Response) {
```
dandwal, line 49: Same as above categories passed to this function but not used

mapper/mapper.go
```diff
@@ -1,203 +1,243 @@
...
+}
+
+func mapSort(page *model.Page, validatedQueryParams data.SearchURLParams, categories []data.Category) {
+	page.Data.Sort.Query = validatedQueryParams.Sort.Query
+
+	page.Data.Sort.LocaliseFilterKeys = validatedQueryParams.Filter.LocaliseKeyName
+
+	page.Data.Sort.LocaliseSortKey = validatedQueryParams.Sort.LocaliseKeyName
+
+	pageSortOptions := []model.SortOptions{}
```
dandwal, line 39: I think it is good practice to declare as a nil slice rather than an empty slice. This might not mean much in some situations however if it is passed to the frontend via json, then `nil` turns into `null`.
```suggestion
	var pageSortOptions []model.SortOptions
```
You can still append to the nil slice e.g.
https://tour.golang.org/moretypes/15

What impact does this have?
The frontend will be given an empty array `[]` in the JSON currently but if we declare this as suggested then if nothing is appended to the slice then the frontend will get `null`.

I am aware this is also done elsewhere so will only raise for new entries in this PR

dandwal: So you can see the impact on the frontend with JS:

Screenshot 2021-03-11 at 11 18 18

Also worth noting this is a frontend usage example, but there could be others who use our APIs as a whole and if this lands in the API then they too will get `[]` rather than `null` once Go converts it to JSON. So always worth declaring slices as nil rather than a literal.

mapper/mapper.go
```diff
@@ -1,203 +1,243 @@
...
-				}
-				itemPage[i].Matches.Description.Title = &matchesTitlePage
-			}
+		item.Matches = &model.Matches{
+			Description: model.MatchDescription{},
+		}
+
+		// Summary Match
+		if matchesDescC.Summary != nil {
+			matchesSummaryPage := []model.MatchDetails{}
```
dandwal, line 155: Same as line 39

mapper/mapper.go
```diff
@@ -1,203 +1,243 @@
...
-						End:   metaDescC.End,
-					})
-				}
-				itemPage[i].Matches.Description.MetaDescription = &matchesMetaDescPage
+			item.Matches.Description.Summary = &matchesSummaryPage
+		}
+
+		// Title Match
+		if matchesDescC.Title != nil {
+			matchesTitlePage := []model.MatchDetails{}
```
dandwal, line 170: Same as line 39

mapper/mapper.go
```diff
@@ -1,203 +1,243 @@
...
-						End:   keywordC.End,
-					})
-				}
-				itemPage[i].Matches.Description.Keywords = &matchesKeywordsPage
+			item.Matches.Description.Title = &matchesTitlePage
+		}
+
+		// Edition Match
+		if matchesDescC.Edition != nil {
+			matchesEditionPage := []model.MatchDetails{}
```
dandwal, line 185: Same as line 39

mapper/mapper.go
```diff
@@ -1,203 +1,243 @@
...
-						End:   datasetIDClient.End,
-					})
-				}
-				itemPage[i].Matches.Description.DatasetID = &matchesDatasetIDPage
+			item.Matches.Description.Edition = &matchesEditionPage
+		}
+
+		// Meta Description Match
+		if matchesDescC.MetaDescription != nil {
+			matchesMetaDescPage := []model.MatchDetails{}
```
dandwal, line 200: Same as line 39

mapper/mapper.go
```diff
@@ -1,203 +1,243 @@
...
+
+			item.Matches.Description.MetaDescription = &matchesMetaDescPage
 		}
 
-		page.Data.Response.Suggestions = respC.Suggestions
-	}
-	page.Data.Response.Items = itemPage
+		// Keywords Match
+		if matchesDescC.Keywords != nil {
+			matchesKeywordsPage := []model.MatchDetails{}
```
dandwal, line 215: Same as line 39

mapper/mapper.go
```diff
@@ -1,203 +1,243 @@
...
-			for _, contentType := range category.ContentTypes {
-				if filter == contentType.Type {
-					filterLocaliseKeyList = append(filterLocaliseKeyList, contentType.LocaliseKeyName)
-				}
+			item.Matches.Description.Keywords = &matchesKeywordsPage
+		}
+
+		// DatasetID Match
+		if matchesDescC.DatasetID != nil {
+			matchesDatasetIDPage := []model.MatchDetails{}
```
dandwal, line 230: Same as line 39

dandwal (commented):
Just some minor comments nothing functionally different

mapper/mapper.go
justinpjose: That is awesome. This makes sense

data/pagination.go
```diff
@@ -0,0 +1,171 @@
...
+
+	return nil
+}
+
+func getLimitFromURLQuery(ctx context.Context, cfg *config.Config, query url.Values) int {
+	defaultLimitStr := strconv.Itoa(cfg.DefaultLimit)
+
+	limit, err := strconv.Atoi(query.Get("limit"))
+	if err != nil {
+		log.Event(ctx, "unable to convert search limit to int - set to default "+defaultLimitStr, log.INFO)
```
daiLlew, line 44: Concatenating strings can get ugly and requires casting of types to string. This gets especially horrific when you are concatenating multiple values. I think a more Go-centric way to achieve the same would be to use `fmt` with placeholders:
```
log.Event(ctx, fmt.Sprintf("unable to convert search limit to int - set to default %d", defaultLimit), log.INFO)
```
If you do this you can remove the string var on line 40.

data/pagination.go
```diff
@@ -0,0 +1,171 @@
...
+
+	return limit
+}
+
+func getPageFromURLQuery(ctx context.Context, cfg *config.Config, query url.Values) int {
+	defaultPageStr := strconv.Itoa(cfg.DefaultPage)
+
+	page, err := strconv.Atoi(query.Get("page"))
+	if err != nil {
+		log.Event(ctx, "unable to convert search page to int - set to default "+defaultPageStr, log.INFO)
```
daiLlew, line 67: same here.

data/pagination.go
```diff
@@ -0,0 +1,171 @@
...
+		query.Set("page", defaultPageStr)
+		page = cfg.DefaultPage
+	}
+
+	if page < 1 {
+		log.Event(ctx, "page number is less than default - default to page "+defaultPageStr, log.INFO)
```
daiLlew, line 73: and here.

data/pagination.go
```diff
@@ -0,0 +1,171 @@
...
+	return page
+}
+
+func getOffset(ctx context.Context, cfg *config.Config, page int, limit int) (offset int, err error) {
+
+	offset = (page - 1) * limit
+
+	// when the offset is negative due to negative current page number or limit
+	if offset < 0 {
+		log.Event(ctx, "offset less than 0 - defaulted to offset "+strconv.Itoa(cfg.DefaultOffset), log.INFO)
```
daiLlew, line 87: same here.

data/pagination.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func getStartPage(currentPage int, totalPages int) int {
+	startPage := currentPage - 2
+
+	if currentPage <= 2 {
+		startPage = 1
+	} else if (currentPage == totalPages-1) || (currentPage == totalPages) {
+		startPage = totalPages - 4
+	}
```
daiLlew, line 139: As someone who has never seen the code before I don't know what the values 2, 1, 4 mean in this context or why they have been chosen. If you create constants for these values with a meaningful name that describes their purpose it makes it clearer to the reader what they are/represent.

data/pagination.go
```diff
@@ -0,0 +1,171 @@
...
+
+	return startPage
+}
+
+func getEndPage(startPage int, totalPages int) int {
+	endPage := startPage + 4
```
daiLlew, line 145: same here.

data/filter.go
```diff
@@ -0,0 +1,186 @@
...
+		if found {
+			validatedQueryParams.Filter.Query = append(validatedQueryParams.Filter.Query, filter.Type)
+			validatedQueryParams.Filter.LocaliseKeyName = append(validatedQueryParams.Filter.LocaliseKeyName, filter.LocaliseKeyName)
+		} else {
+			err := errs.ErrFilterNotFound
+			logData := log.Data{"filter not found": filter}
+			log.Event(ctx, "failed to find filter", log.Error(err), log.ERROR, logData)
+
+			return err
+		}
```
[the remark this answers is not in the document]
justinpjose: I completely agree. Nice

data/pagination.go
justinpjose: Of course. That makes complete sense - the little things are really important and I appreciate that

handlers/handlers.go
```diff
@@ -28,86 +27,165 @@ type RenderClient interface {
...
+	for _, responseType := range countResp.ContentTypes {
+		foundFilter := false
+
+	categoryLoop:
+		for i, category := range categories {
+			for j, contentType := range category.ContentTypes {
+				for _, subType := range contentType.SubTypes {
+					if responseType.Type == subType {
+						categories[i].Count += responseType.Count
+						categories[i].ContentTypes[j].Count += responseType.Count
```
daiLlew, line 130: You need to be careful when accessing array element by index.  I think this specific case is OK because `categories` is an array of **non pointer** types so each element of the array cannot be `nil`. No action to take here but I thought it would highlight the potential risk so you are aware of it.

If categories was a list of pointer types i.e. `[]*data.Category` you would need to be careful and not assume an entry is not nil.

For example if you had in the input
```
[]*data.Category{nil, nil, nil}
```
The `for` loop would iterate over the array entries but it would nil pointer error when it tries to do `categories[i].Count`.

This is essentially what happened in last weeks issue and what we discussed on the postmortem call. The simple solution to this is to _alway assume it could be nil_ and do a nil check before you access it.

daiLlew (changes requested):

data/filter.go
```diff
@@ -0,0 +1,185 @@
...
+// ContentType represents the type of the search results and the number of results for each type
+type ContentType struct {
+	LocaliseKeyName string   `json:"localise_key"`
+	Count           int      `json:"count"`
+	Type            string   `json:"type"`
+	SubTypes        []string `json:"sub_types"`
+}
+
+// Categories represent the list of all search categories
+var Categories = []Category{Publication, Data, Other}
```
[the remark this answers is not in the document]
justinpjose: Love this

data/filter_test.go
```diff
@@ -0,0 +1,167 @@
...
+			Convey("Then return no error", func() {
+				So(err, ShouldBeNil)
+			})
+
+			Convey("And update validatedQueryParams for filter", func() {
+				So(validatedQueryParams, ShouldNotBeNil)
+				So(validatedQueryParams.Filter.Query, ShouldResemble, []string{"article"})
+				So(validatedQueryParams.Filter.LocaliseKeyName, ShouldResemble, []string{"Article"})
+			})
+		})
```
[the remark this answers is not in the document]
justinpjose: This is awesome. I never thought about these test scenarios so thank you so so much

data/pagination.go
```diff
@@ -0,0 +1,177 @@
...
+		log.Event(ctx, fmt.Sprintf("unable to convert search page to int - set to default %d", cfg.DefaultPage), log.INFO)
+		query.Set("page", strconv.Itoa(cfg.DefaultPage))
+		page = cfg.DefaultPage
+	}
+
+	if page < 1 {
+		log.Event(ctx, fmt.Sprintf("page number is less than default - default to page %d", cfg.DefaultPage), log.INFO)
+		query.Set("page", strconv.Itoa(cfg.DefaultPage))
+		page = cfg.DefaultPage
+	}
```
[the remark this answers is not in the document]
justinpjose: There isn't a maximum page limit. The total number of pages is calculated from the number of search results retrieved and the limit set. There is a maximum number of search results (`DefaultMaximumSearchResults: 500`) and a maximum limit (`DefaultMaximumLimit: 50`) which in turn will put a limit on the calculated number of pages

data/query_test.go
```diff
@@ -0,0 +1,206 @@
...
+			"page":   []string{"1"},
+		}
+
+		Convey("When ReviewQuery is called", func() {
+			_, err := ReviewQuery(ctx, cfg, urlQuery)
+
+			Convey("Then return an error", func() {
+				So(err, ShouldNotBeNil)
+			})
+		})
```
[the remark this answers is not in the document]
justinpjose: If the query is `empty`, it will retrieve `all` the search results available (up to a maximum of `500` results)
If the query is `invalid`, the `search-api` would retrieve `0` search results

handlers/handlers.go
```diff
@@ -28,86 +27,165 @@ type RenderClient interface {
...
+		log.Event(ctx, "unable to review query", log.Error(err), log.ERROR)
+		setStatusCode(w, req, err)
+		return
+	}
+
+	apiQuery := data.GetSearchAPIQuery(validatedQueryParams)
+
+	resp, err := searchC.GetSearch(ctx, apiQuery)
+	if err != nil {
+		log.Event(ctx, "getting search response from client failed", log.Error(err), log.ERROR)
```
daiLlew, line 56: I think it would be useful to include the query in `log.Data`

daiLlew (approved):

data/pagination.go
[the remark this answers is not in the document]
nshumoogum: @daiLlew - we need to limit the user continuously paging through search results as it results in slower and slower responses from the search API due to how an elasticsearch cluster is setup when there is more than 1 node.

Generally it is bad practice to indefinitely page results due to causing performance issues. The performance knock usually depends on how the database in question is able to cursor through the data. Elasticsearch actually has it's own hard limit of 10,000 results.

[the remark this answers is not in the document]
nshumoogum: In the context of searching for data, the user should be able to find what they are looking for in the top 5 results (maybe top 100 for the ONS sitewide search), the limits we have set will allow the first 500 results which is more than enough results. If the user cannot find what they are looking for in the top 500 then we have made a bad search engine or we don't have the data.

data/pagination.go
```diff
@@ -0,0 +1,177 @@
...
+	return page
+}
+
+func getOffset(ctx context.Context, cfg *config.Config, page int, limit int) (offset int, err error) {
+
+	offset = (page - 1) * limit
+
+	// when the offset is negative due to negative current page number or limit
+	if offset < 0 {
+		log.Event(ctx, fmt.Sprintf("offset less than 0 - defaulted to offset %d", cfg.DefaultOffset), log.INFO)
```
nshumoogum, line 87: This should probably a WARN log line - if we have many of these, it could help us identify someone is doing something malicious to us. If it stayed as a INFO log, more than likely we will ignore this.

handlers/handlers.go
```diff
@@ -28,86 +27,170 @@ type RenderClient interface {
...
+		return
 	}
-	log.Event(req.Context(), "setting-response-status", log.Error(err), log.ERROR)
-	w.WriteHeader(status)
 }
 
-var marshal = func(v interface{}) ([]byte, error) {
-	return json.Marshal(v)
+// validateCurrentPage checks if the current page exceeds the total pages which is a bad request
+func validateCurrentPage(ctx context.Context, validatedQueryParams data.SearchURLParams, resultsCount int) (err error) {
```
nshumoogum, line 87: I wouldn't initialise err in function signature, so:

`func validateCurrentPage(ctx context.Context, validatedQueryParams data.SearchURLParams, resultsCount int) error {`

Then on line 93 add `:` to set err variable

That way it is following the same pattern as your other functions in this package.

nshumoogum (changes requested):
Looking good just some minor comments

data/pagination.go
```diff
@@ -0,0 +1,177 @@
...
+		pagesToDisplay = append(pagesToDisplay, model.PageToDisplay{
+			PageNumber: i,
+			URL:        getPageURL(query, i, controllerQuery),
+		})
+	}
+
+	return pagesToDisplay
+}
+
+func getStartPage(cfg *config.Config, currentPage int, totalPages int) int {
```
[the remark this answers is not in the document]
justinpjose: Yes. That's exactly it. And your example is correct as well because we are only showing 5 available pages at the bottom. So if we are on page 5, then it would be 3,4,5,6,7

handlers/handlers.go
```diff
@@ -28,86 +27,170 @@ type RenderClient interface {
...
+
+	resp, err := searchC.GetSearch(ctx, apiQuery)
+	if err != nil {
+		logData := log.Data{"api query passed to search-api": apiQuery}
+		log.Event(ctx, "getting search response from client failed", log.Error(err), log.ERROR, logData)
+		setStatusCode(w, req, err)
+		return
+	}
+
+	// TO-DO: Until API handles aggregration on datatypes (e.g. bulletins, article), we need to make a second request
```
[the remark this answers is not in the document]
justinpjose: The second request is carried out in the function `getCategoriesTypesCount` (in `handlers.go` line `104`). 
We cannot remove the comment because right now, the number of search results for `every` filter type is retrieved by doing a second request to the API (in the second request, we are not passing any filters so the search-API provides all the possible search results with count for every type).  

The first request only provides the number of search results for `each` filter type applied by the user

In the future, we want the API to handle aggregation on filter datatypes where we can avoid the second request from controller to api

nshumoogum (approved):

PR #18, opened 2021-02-23
Release 0.4.0

cadmiumcat:
### What

Updates version of `dp-api-clients-go` to v1.33.4. This version which contains a name change (from `dp-query` to `dp-search`). This affects the healthcheck message

### How to review

Check that the correct commits have been included

### Who can review

Anyone


mike-kelly-methods (approved):
👍

PR #23, opened 2021-05-25
separate main function to new world layout

justinpjose:
### What
- Separate main function to new world layout (which now contains a `Run` function too)
- This work is based on the code layout in the `dp-frontend-filter-dataset-controller`

It might look like a lot of changes but it isn't. Most of them are new mocks generated by moq.
This work is part of updating the controller to healthcheck against API router. But before updating the controller, the main function needed to be updated so that it is up to date with the layout of the new con 

### How to review
- Check if `make debug` works as expected
- Check if the tests pass (`make test`)

### Who can review
Anyone


Marianasilvago:
Should we be committing the binary file in the repository.

handlers/clients.go
```diff
@@ -0,0 +1,26 @@
...
+
+	searchC "github.com/ONSdigital/dp-api-clients-go/site-search"
+)
+
+//go:generate moq -out ../mocks/clients.go -pkg mocks . RenderClient SearchClient
+
+// ClientError is an interface that can be used to retrieve the status code if a client has errored
+type ClientError interface {
+	Error() string
+	Code() int
```
ChrysmOre, line 15: I understand this may be copied from somewhere else but I really think StatusCode() is more appropriate than Code() for retrieving the status code.. It falls in line with implementations found elsewhere in the Go world and is more descriptive.. 'Code' could mean 'Exit Code' for example.

ChrysmOre: Furthermore - I can't see where this is being used? But I do remember seeing this implementation somewhere else (where it is used to assert an error check it's statusCode) and remember noticing that it was actually unnecessary to have the full error as an interface - the interface just needs to be for Code() (or StatusCode() :p) function to ascertain the statusCode. With interfaces it's better to include only the behaviour you need.

main.go
```diff
@@ -4,92 +4,60 @@ import (
...
-}
+	log.Event(ctx, "got service configuration", log.Data{"config": cfg})
 
-func registerCheckers(ctx context.Context, h *health.HealthCheck, c routes.Clients) (err error) {
-	if err = h.AddCheck("frontend renderer", c.Renderer.Checker); err != nil {
-		log.Event(ctx, "failed to add frontend renderer checker", log.Error(err))
+	// Run service
+	srv, err := service.Run(ctx, cfg, svcList, svcErrors)
+	if err != nil {
+		log.Event(ctx, "failed to run service", log.Error(err))
```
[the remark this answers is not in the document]
ChrysmOre: aren't there functions like 'log.Error, log.Warn' etc which should probably be preferably used? (less chance of mistake, shorter lines)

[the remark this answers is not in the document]
nshumoogum: Yes but I would suggest we write a script to convert all logs in an app to use the latest log library and to use the new format as you suggested.

Arguably not having the script to do this and by implementing the `log.Error(...)` etc. here now would (likely) not result in any issues running a script later.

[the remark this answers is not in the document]
nshumoogum: Maybe once the script has been written, a demo on the changes to writing our app logs using log.go lib; one for the guild?

handlers/clients.go
justinpjose: Yeah I agree. This is a very good point. It should be specific. I can't change the name to StatusCode() as I would need to change it in the `dp-api-clients-go` which means refactoring all the apps that use the renderer client for example. This would require a lot of changes throughout apps

nshumoogum: I just wanted to add to what Justin has said and it would make more sense to update the library `dp-api-clients-go` before rolling out a change to any apps regarding retrieval of status code; but agree `StatusCode()` would be better.

nshumoogum (approved):

nshumoogum:
Missing new configuration variable in README.md - `API_ROUTER_URL`

nshumoogum (approved):

PR #24, opened 2021-06-08
Release/0.6.0

justinpjose:
### What
- Separate main function to new world layout (which now contains a `Run` function too)
- This work is based on the code layout in the `dp-frontend-filter-dataset-controller`
- And healthcheck against API router

### How to review
- Please note that this release only contains commits from this PR https://github.com/ONSdigital/dp-frontend-search-controller/pull/23

### Who can review
Anyone


cookel2 (commented):
@justinpjose I think it would be a good idea to run 'go mod tidy' as I tried it locally and it made quite a few changes to the go.sum file.

handlers/clients.go
```diff
@@ -0,0 +1,25 @@
...
+	"net/url"
+
+	searchC "github.com/ONSdigital/dp-api-clients-go/site-search"
+)
+
+//go:generate moq -out clients_mock.go -pkg handlers . RenderClient SearchClient
+
+// ClientError is an interface that can be used to retrieve the status code if a client has errored
+type ClientError interface {
+	Code() int
```
cookel2, line 14: I had a look at the handlers_test and it mocks this interface but includes the Error() method, which seems to be missing from here? 
```suggestion
	Error() string
	Code() int
```

service/service_test.go
```diff
@@ -0,0 +1,436 @@
...
+				Config:      cfg,
+				HealthCheck: hcMock,
+				Server:      failingServerCloseMock,
+				ServiceList: serviceList,
+			}
+
+			Convey("When closing the service", func() {
+				err = svc.Close(ctx)
+
+				Convey("Then Close operation tries to close all dependencies and returns an error", func() {
```
cookel2, line 363: The "and returns an error" part is done in the following Convey so should probably be removed from this one
```suggestion
				Convey("Then Close operation tries to close all dependencies", func() {
```

cookel2 (commented):
I've added a couple more comments @justinpjose

handlers/clients.go
justinpjose: We are only using the `ClientError` (without mocking) in the function `setStatusCode` (https://github.com/ONSdigital/dp-frontend-search-controller/blob/develop/handlers/handlers.go#L165). In that function, we are using `ClientError` to map the error to get the `Code` of the error. We don't care about the Error() when it comes to the interface itself.

justinpjose: This was discussed in the PR (https://github.com/ONSdigital/dp-frontend-search-controller/pull/23). I have unresolved the particular discussion so you can see it straight away

cookel2 (approved):
looks good to me

cookel2 (approved):

PR #27, opened 2021-06-17
Fix pagination showing more than default max results

jondewijones:
### What

Fix pagination showing more than default max results

### How to review

While running new search you should no longer see pages and result beyond default max result currently set at 500. 
e.g if you have results at 10 per page you should only be able to get 50 pages no more, or if set at 50 per page then only 10 pages

### Who can review

Anyone

### DIFF
--- a/data/pagination.go
+++ b/data/pagination.go
@@ -102,7 +102,10 @@ func getOffset(ctx context.Context, cfg *config.Config, page int, limit int) (of
 }
 
 // GetTotalPages gets the total pages of the search results
-func GetTotalPages(limit int, count int) int {
+func GetTotalPages(cfg *config.Config, limit, count int) int {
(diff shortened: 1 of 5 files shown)


data/pagination.go
```diff
@@ -102,7 +102,10 @@ func getOffset(ctx context.Context, cfg *config.Config, page int, limit int) (of
 }
 
 // GetTotalPages gets the total pages of the search results
-func GetTotalPages(limit int, count int) int {
+func GetTotalPages(cfg *config.Config, limit, count int) int {
```
daiLlew, line 105: Change looks fine but you should add a new test case to cover this updated behaviour.

jondewijones: 👍

daiLlew (approved):

PR #28, opened 2021-06-21
Remove search api url config property

jondewijones:
### What

Remove search api url config property

### Who can review

Anyone


sandypadmanabhan (approved):
LGTM

PR #31, opened 2021-07-01
add component test for health endpoint

justinpjose:
### What
- Adding component test for health endpoint

### How to review
- Run `make test-component`

### Who can review
Anyone


nshumoogum (changes requested):

features/healthcheck.feature
```diff
@@ -0,0 +1,171 @@
...
+                        "last_success": null,
+                        "last_failure": "0001-01-01T00:00:00Z"
+                    }
+                ]
+            }
+        """
+
+    Scenario: Returning a CRITICAL (500) status when health endpoint called
+        Given one the downstream services is failing
+        And I wait "2" seconds for the healthcheck to be available
```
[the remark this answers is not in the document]
justinpjose: When I do this, the component test fails and returns a `429` response instead of `500`

features/steps/component.go
```diff
@@ -0,0 +1,159 @@
...
+
+	return nil
+}
+
+// InitialiseService returns the http.Handler that's contained within the component.
+func (c *Component) InitialiseService() (http.Handler, error) {
+	return c.HTTPServer.Handler, nil
+}
+
+func getHealthCheckOK(cfg *config.Config, buildTime, gitCommit, version string) (service.HealthChecker, error) {
```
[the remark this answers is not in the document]
justinpjose: The function signature needs to have them so that it can satisfy the interface of the HealthCheck function

features/healthcheck.feature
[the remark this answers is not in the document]
justinpjose: We have possibly discovered a bug in the `dp-healthcheck` library where this will fail if we don't make two requests to `/health`.

ci/build.yml
```diff
@@ -6,7 +6,7 @@ image_resource:
   type: docker-image
   source:
     repository: golang
-    tag: 1.16.1
+    tag: 1.16.6
```
redhug1, line 9: shouldn't this be just 1.16 and then the latest sub version is pulled in ?

ci/component.yml
```diff
@@ -6,7 +6,7 @@ image_resource:
   type: docker-image
   source:
     repository: golang
-    tag: 1.16.1
+    tag: 1.16.6
```
redhug1, line 9: shouldn't this be just 1.16 and then the latest sub version is pulled in ?

ci/unit.yml
```diff
@@ -6,7 +6,7 @@ image_resource:
   type: docker-image
   source:
     repository: golang
-    tag: 1.16.1
+    tag: 1.16.6
```
redhug1, line 9: shouldn't this be just 1.16 and then the latest sub version is pulled in ?

redhug1 (changes requested):
I ran the linter against dp-frontend-search-controller and it shows these issues, that are easy to fix:
routes/routes.go:16:1: commentFormatting: put a space between `//` and comment text (gocritic)
//Clients - struct containing all the clients for the controller
^
handlers/handlers.go:117:2: commentFormatting: put a space between `//` and comment text (gocritic)
	//Remove filter to get count of all types for the query from the client
	^
handlers/handlers_test.go:518:28: ineffectual assignment to err (ineffassign)
		mockCountSearchResponse, err := mapper.GetMockSearchResponse()
		                         ^
handlers/handlers_test.go:564:28: ineffectual assignment to err (ineffassign)
		mockCountSearchResponse, err := mapper.GetMockSearchResponse()
		                         ^
mapper/mapper_test.go:44:10: ineffectual assignment to err (ineffassign)
		respC, err := GetMockSearchResponse()
		       ^

i find that i have to keep running the linter after i fix things as it then reveals more of the same issues (a bit annoying).

The linter line i use is:

golangci-lint --deadline=10m --fast --enable=gosec --enable=gocritic --enable=gofmt --enable=gocyclo --enable=bodyclose --enable=gocognit run

-=-=-

Make test-component   is reporting failures ... needs fixing.

ci/build.yml
justinpjose: When looking at all the other repos, the patch number is also given. For example; https://github.com/ONSdigital/dp-topic-api/blob/develop/ci/build.yml#L9

ci/component.yml
justinpjose: When looking at all the other repos, the patch number is also given. For example; https://github.com/ONSdigital/dp-topic-api/blob/develop/ci/build.yml#L9

ci/unit.yml
justinpjose: When looking at all the other repos, the patch number is also given. For example; https://github.com/ONSdigital/dp-topic-api/blob/develop/ci/build.yml#L9

redhug1 (approved):
LGTM

nshumoogum (approved):

PR #34, opened 2021-07-22
Use auth if present when in publishing

jondewijones:
### What

- Use auth if present when in publishing
- Update api clients to v2

### Who can review

Anyone


morses-code (commented):
One tiny thing, `go mod tidy` to remove the previous `dp-api-clients-go` lib.

Otherwise, LGTM

handlers/handlers_test.go
```diff
@@ -411,13 +411,13 @@ func TestUnitGetCategoriesTypesCountFailure(t *testing.T) {
...
 		mockedSearchClient := &SearchClientMock{
-			GetSearchFunc: func(ctx context.Context, query url.Values) (searchC.Response, error) {
+			GetSearchFunc: func(ctx context.Context, userAuthToken, serviceAuthToken, collectionID string, query url.Values) (searchC.Response, error) {
 				return searchC.Response{}, errs.ErrInternalServer
 			},
 		}
 
 		Convey("When getCategoriesTypesCount is called", func() {
-			categories, err := getCategoriesTypesCount(ctx, apiQuery, mockedSearchClient)
+			categories, err := getCategoriesTypesCount(ctx, accessToken, collectionID, apiQuery, mockedSearchClient)
```
daiLlew, line 420: We should update each test that has been affect by this change and add a check to verify the client is being called with the `accessToken`  and `collectionID`.

```
So(mockedSearchClient.GetSearchFuncCalls[0].userAuthToken, ShouldEqual, expectedToken)
So(mockedSearchClient.GetSearchFuncCalls[0].collectionID, ShouldEqual, expectedID)
```

This should hopefully by copy past into the appropriate test cases.

daiLlew (commented):
The change looks fine to me but I think we should update the tests to prove the changes introduced - see comment for more details.

daiLlew (approved):

PR #41, opened 2021-11-02
Move models into controller

jondewijones:
### What

Move models into controller. Use internal models rather dp-frontend-models 

### Who can review

Anyone

### DIFF
--- a/model/search.go
+++ b/model/search.go
@@ -1,0 +1,123 @@
+package model
+
+import "github.com/ONSdigital/dp-renderer/model"
+
+// Search is the model struct for the cookies preferences form
+type SearchPage struct {
+	model.Page
+	Data       Search      `json:"data"`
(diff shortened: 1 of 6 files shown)


model/search.go
```diff
@@ -0,0 +1,123 @@
...
+	Count           int    `json:"count"`
+	LocaliseKeyName string `json:"localise_key"`
+}
+
+// ContentItem represents each search result
+type ContentItem struct {
+	Description Description `json:"description"`
+	Type        string      `json:"type"`
+	URI         string      `json:"uri"`
+	Matches     *Matches    `json:"matches,omitempty"`
```
rav-pradhan, line 63: Could we replace this with `MatchDescriptions` and remove a level of nesting, as the `Matches` struct seems to only include that?

rav-pradhan (approved):

PR #47, opened 2021-11-25
add templating to id and for

NoelleDL:
### What

Checkboxes had non-unique values so they could only select the first checkbox. Added templating to variables so that `id` and `for` values are unique. 

### How to review

On new search results page on the left under `Content types to show`, see that one can now choose multiple checkboxes. Inspect code and see that `id` and `for` should be the selected content type and not `$id`.  
Screenshot 2021-11-25 at 08 50 48

### Who can review

Anyone but me.

### DIFF
--- a/assets/templates/partials/filter.tmpl
+++ b/assets/templates/partials/filter.tmpl
@@ -20,8 +20,8 @@
                                     <li class="filters__item">
                                         <div class="filters__field">
                                             {{ $id := (print "checkout-" $content.Type) }}
-                                            <input id=$id class="js-auto-submit__input" type="checkbox" name="filter" value={{ $content.Type }} {{ range $filter }}{{ if eq . $content.Type }}checked{{end}}{{end}}>
-                                            <label for=$id>
+                                            <input id={{ $id }} class="js-auto-submit__input" type="checkbox" name="filter" value={{ $content.Type }} {{ range $filter }}{{ if eq . $content.Type }}checked{{end}}{{end}}>
+                                            <label for={{ $id }}>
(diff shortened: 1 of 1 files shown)


assets/templates/partials/filter.tmpl
```diff
@@ -20,8 +20,8 @@
                                     <li class="filters__item">
                                         <div class="filters__field">
                                             {{ $id := (print "checkout-" $content.Type) }}
-                                            <input id=$id class="js-auto-submit__input" type="checkbox" name="filter" value={{ $content.Type }} {{ range $filter }}{{ if eq . $content.Type }}checked{{end}}{{end}}>
-                                            <label for=$id>
+                                            <input id={{ $id }} class="js-auto-submit__input" type="checkbox" name="filter" value={{ $content.Type }} {{ range $filter }}{{ if eq . $content.Type }}checked{{end}}{{end}}>
```
mr-nick17, line 23: You need to add the quotation marks
```suggestion
                                            <input id="{{ $id }}" class="js-auto-submit__input" type="checkbox" name="filter" value={{ $content.Type }} {{ range $filter }}{{ if eq . $content.Type }}checked{{end}}{{end}}>
```

assets/templates/partials/filter.tmpl
```diff
@@ -20,8 +20,8 @@
...
+                                            <label for={{ $id }}>
```
mr-nick17, line 24: Same here
```suggestion
                                            <label for="{{ $id }}">
```

mr-nick17 (approved):
LGTM

assets/templates/partials/filter.tmpl
NoelleDL: Ah, thanks!

PR #54, opened 2021-12-02
Fix/show correct data type

NoelleDL:
### What

Search results were sometimes returning data `subTypes` (eg, `article_download`) which did not match the selected filter option. Search results now return the correct data type that matches the selected filter option. The data type showing is the singular instead of the plural version of the selected filter option. 

### How to review

Go to new search. Search for something (eg, "economics"), select filter options. See that singular word appears that matches the filter option.

### Who can review

Anyone but me.

### DIFF
--- a/assets/locales/service.cy.toml
+++ b/assets/locales/service.cy.toml
@@ -54,10 +54,12 @@ other = "Compendiwm"
 
 [TimeSeries]
 description = "Time series"
+one = "Cyfres amser"
 other = "Cyfres amser"
 
 [Datasets]
 description = "Datasets"
+one = "Set ddata"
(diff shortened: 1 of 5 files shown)


assets/locales/service.cy.toml
```diff
@@ -71,10 +73,12 @@ one = "Dyddiad y datganiad"
 
 [UserRequestedData]
 description = "User requested data"
+one = "Data a gyrchwyd gan y defnyddiwr"
```
jondewijones, line 76: we'll need to replicate this translation in the English file as well

jondewijones (approved):
LGTM

PR #65, opened 2022-01-10
use default list of content types if no filters are selected. prevent unwanted indexed content types being returned in search

jondewijones:
### What

use default list of content types if no filters are selected. prevent unwanted indexed content types being returned in search

### Who can review

Anyone

### DIFF
--- a/data/filter.go
+++ b/data/filter.go
@@ -30,6 +30,27 @@ type ContentType struct {
 	SubTypes        []string `json:"sub_types"`
 }
 
+var defaultContentTypes = "article," +
+	"article_download," +
+	"bulletin," +
+	"compendium_landing_page," +
+	"compendium_chapter," +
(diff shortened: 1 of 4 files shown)


data/filter.go
```diff
@@ -30,6 +30,27 @@ type ContentType struct {
 	SubTypes        []string `json:"sub_types"`
 }
 
+var defaultContentTypes = "article," +
```
IncredibleMrTim, line 33: this would be cleaner as a string literal

jondewijones: done like this so it was easier to add/remove/maintain the list going forward and so it was consistent with search-api https://github.com/ONSdigital/dp-search-api/blob/develop/api/search.go#L17

IncredibleMrTim (approved):

PR #71, opened 2022-01-13
Add new content types mapping

jondewijones:
### What

Add new content types mapping

### Who can review

Anyone

### DIFF
--- a/data/filter.go
+++ b/data/filter.go
@@ -101,7 +101,7 @@ var (
 	Datasets = ContentType{
 		LocaliseKeyName: "Datasets",
 		Type:            "datasets",
-		SubTypes:        []string{"dataset_landing_page", "timeseries_dataset", "reference_tables"},
+		SubTypes:        []string{"dataset_landing_page", "timeseries_dataset"},
 	}
 
 	// UserRequestedData - Search information specific for user requested data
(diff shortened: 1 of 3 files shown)


data/filter.go
```diff
@@ -129,6 +125,12 @@ var (
 		SubTypes:        []string{"static_foi", "static_page", "static_landing_page", "static_article"},
 	}
 
+	ProductPage = ContentType{
+		LocaliseKeyName: "ProductPage",
+		Type:            "product_page",
+		SubTypes:        []string{},
```
justinpjose, line 131:
```suggestion
		SubTypes:        []string{"product_page"},
```

data/filter.go
```diff
@@ -105,7 +101,7 @@ var (
 	Datasets = ContentType{
 		LocaliseKeyName: "Datasets",
 		Type:            "datasets",
-		SubTypes:        []string{"dataset_landing_page", "reference_tables"},
+		SubTypes:        []string{"dataset_landing_page", "timeseries_dataset", "reference_tables"},
```
justinpjose, line 104: As you removed `reference_tables` from `defaultContentTypes`, you need to remove it from here as well. This is in the assumption that we don't consider search results with the type `reference_tables` or they just don't exist

justinpjose (approved):

PR #76, opened 2022-01-17
update type field and move loop logic to controller

justinpjose:
### What
1. Change the field names in `ContentType` struct from `Type` -> `Group` and `SubTypes` -> `Types`. This gives clarity on what each field means. This is an example of how each field links to each other:

```
Publication (Category)
-> Article (Group)
-> -> "articles","static_articles" (Types)
```

2. Moved the logic to get the search result `Type` `LocaliseKey` from the template to the controller 

### How to review
- Check if the code changes makes sense
- Run locally and see if the search page is functioning as normal
- Check if the tests pass

### Who can review
Anyone

### DIFF
--- a/data/filter.go
+++ b/data/filter.go
@@ -26,8 +26,8 @@ type Category struct {
 type ContentType struct {
 	LocaliseKeyName string   `json:"localise_key"`
 	Count           int      `json:"count"`
-	Type            string   `json:"type"`
-	SubTypes        []string `json:"sub_types"`
+	Group           string   `json:"group"`
+	Types           []string `json:"types"`
 }
 
 var defaultContentTypes = "article," +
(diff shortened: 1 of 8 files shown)


NoelleDL (dismissed):

data/filter.go
```diff
@@ -72,75 +72,76 @@ var (
...
+		Group:           "bulletin",
+		Types:           []string{"bulletin"},
 	}
 
 	// Article - Search information specific for articles
 	Article = ContentType{
 		LocaliseKeyName: "Article",
-		Type:            "article",
-		SubTypes:        []string{"article", "article_download"},
+		Group:           "article",                               //Type
```
nshumoogum, line 82: Probably need to remove comments on this line and the one below

data/filter.go
```diff
@@ -72,75 +72,76 @@ var (
...
+		Types:           []string{"timeseries"},
 	}
 
 	// Datasets - Search information specific for datasets
 	Datasets = ContentType{
 		LocaliseKeyName: "Datasets",
-		Type:            "datasets",
-		SubTypes:        []string{"dataset_landing_page", "timeseries_dataset"},
+		Group:           "datasets",
+		Types:           []string{"dataset_landing_page", "reference_tables"},
```
nshumoogum, line 104: Im not sure why this has changed back to `reference_tables` from `timeseries_dataset`

handlers/handlers.go
```diff
@@ -134,7 +134,7 @@ func setCountToCategories(ctx context.Context, countResp searchCli.Response, cat
 	categoryLoop:
 		for i, category := range categories {
 			for j, contentType := range category.ContentTypes {
-				for _, subType := range contentType.SubTypes {
+				for _, subType := range contentType.Types {
```
nshumoogum, line 137: Probably should remove empty file `helll` ;)

nshumoogum (changes requested):

data/filter.go
justinpjose: Done

data/filter.go
justinpjose: No idea either but this has been updated

handlers/handlers.go
justinpjose: Sorry about that

nshumoogum (approved):

PR #83, opened 2022-02-03
Feature/ds pagination sort and filter

liamjlewis:
### What

This work targets all of the acceptance criteria here:
https://trello.com/c/E06s8Zta/882-design-system-implemented-on-new-search

The markup has been changed to allow for proper styling, javascript targeting and non-JS use.

Changing filters, sort and pagination will dynamically alter the article results, pagination and summary info and the query in the URL (allowing users to bookmark and manually share links)

### How to review
Ideally view the changes locally since they are mostly visual and will require manual testing.

Also (as stated above) this work is dependant on my changes to the dp-design-system, so please have this branch running locally:
https://github.com/ONSdigital/dp-design-system/tree/feature/DS-pagination-sort-and-filter

This type definition is also required to run locally.
https://github.com/ONSdigital/dp-renderer/pull/53

### Who can review
Anyone.

### DIFF
--- a/assets/templates/partials/filter.tmpl
+++ b/assets/templates/partials/filter.tmpl
@@ -26,46 +26,50 @@
             <a href="/search{{ if .Data.Query }}?q={{ .Data.Query}}{{end}}" id="clear-search" class="float-right font-size--18">{{ localise "ClearAll" $lang 1 }}</a>
             <p class="ons-checkboxes__label">Show only:</p>
             {{ range $index, $topFilter := $filters}}
-                <div class="ons-checkboxes__items">
-                    <span class="ons-checkboxes__item ons-checkboxes__item--no-border">
-                        <span class="ons-checkbox ons-checkbox--no-border category-filter">
-                            <input type="checkbox" id="group-{{ $index }}"
-                                class="ons-checkbox__input ons-js-checkbox"
-                                value="{{ print $topFilter.FilterKey }}" categoryChildren="{{ print $topFilter.FilterKey }}" aria-controls="group-{{ $index }}-other-wrap"
-                                aria-haspopup="true"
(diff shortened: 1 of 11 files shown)


assets/templates/partials/filter.tmpl
```diff
@@ -1,83 +1,94 @@
-{{ $filter := .Data.Filter }}
+{{ $filters := .Data.Filters }}
 {{ $response := .Data.Response }}
 {{ $query := .Data.Query }}
+{{ $sort := .Data.Sort.Query }}
 {{ $lang := .Language }}
-<form id="filterForm" class="js-auto-submit__form" action="/search" method="get">
+<form id="filterForm" method="get">
```
jondewijones, line 6: Think this change may break non-js version as it will no longer submit to server?

assets/templates/partials/filter.tmpl
```diff
@@ -1,83 +1,94 @@
...
-                                                               value="{{ $id }}"
-                                                        />
-                                                        <label class="ons-checkbox__label" for="{{ $id }}"
-                                                               id="{{ $id }}-label">
-                                                            {{ localise $content.LocaliseKeyName $lang 4 }} ({{ $content.Count }})
-                                                        </label>
-                                                    </span>
+            </h2>
+            <span class="ons-collapsible__icon">
+                <svg class="ons-svg-icon " viewBox="0 0 8 13" xmlns="http://www.w3.org/2000/svg" focusable="false" fill="currentColor">
```
jondewijones, line 19: We should probably add this icon as a template in the renderer to be consistent? 

See: https://github.com/ONSdigital/dp-renderer/tree/main/assets/templates/icons

jondewijones:
image
Could possibly do with some additional whitespace space between pagination and feedback banner?

PR #89, opened 2022-02-21
Number of results

liamjlewis:
https://trello.com/c/D5CEhvOF/962-number-of-results

### What

If the user uses an empty string they are sent to the search page and no API call is made.

### How to review

The usual, either reading the code or pulling and checking it visually.

### Who can review

Anyone.

### DIFF
--- a/handlers/handlers.go
+++ b/handlers/handlers.go
@@ -28,7 +28,7 @@ func read(w http.ResponseWriter, req *http.Request, cfg *config.Config, rend Ren
 
 	urlQuery := req.URL.Query()
 
-	validatedQueryParams, err := data.ReviewQuery(ctx, cfg, urlQuery)
+	validatedQueryParams, validationProblem, err := data.ReviewQuery(ctx, cfg, urlQuery)
 	if err != nil {
 		log.Error(ctx, "unable to review query", err)
 		setStatusCode(w, req, err)
@@ -40,6 +40,15 @@ func read(w http.ResponseWriter, req *http.Request, cfg *config.Config, rend Ren
 	var searchResp searchCli.Response
 	var respErr error
 	var departmentResp searchCli.Department
+
+	if validationProblem {
+		// avoid making any API calls
+		basePage := rend.NewBasePageModel()
+		m := mapper.CreateSearchPage(cfg, req, basePage, validatedQueryParams, []data.Category{}, searchResp, departmentResp, lang, validationProblem)
+		rend.BuildPage(w, m, "search")
+		return
+	}
+
 	var wg sync.WaitGroup
 	wg.Add(2)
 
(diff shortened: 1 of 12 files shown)


handlers/handlers.go
```diff
@@ -28,7 +28,7 @@ func read(w http.ResponseWriter, req *http.Request, cfg *config.Config, rend Ren
 
 	urlQuery := req.URL.Query()
 
-	validatedQueryParams, err := data.ReviewQuery(ctx, cfg, urlQuery)
+	validatedQueryParams, validationProblem, err := data.ReviewQuery(ctx, cfg, urlQuery)
```
nshumoogum, line 31: So instead of handling boolean field, check the error and then make the decision on what to do; something like:

```
        validatedQueryParams, err := data.ReviewQuery(ctx, cfg, urlQuery)
	if err != nil && err != apperrors.<your new error <- needs to be added to apperrors package> {
		log.Error(ctx, "unable to review query", err)
		setStatusCode(w, req, err)
		return
	}
```

Then where you check validateProblem below, change that to `if err == apperrors.<your new error <- needs to be added to apperrors package> {`

See next comment

handlers/handlers.go
```diff
@@ -40,6 +40,15 @@ func read(w http.ResponseWriter, req *http.Request, cfg *config.Config, rend Ren
 	var searchResp searchCli.Response
 	var respErr error
 	var departmentResp searchCli.Department
+
+	if validationProblem {
```
nshumoogum, line 44:
```suggestion
	if err ==  apperrors.<your new error <- needs to be added to apperrors package> {
```

handlers/handlers.go
```diff
@@ -40,6 +40,15 @@ func read(w http.ResponseWriter, req *http.Request, cfg *config.Config, rend Ren
...
+		// avoid making any API calls
+		basePage := rend.NewBasePageModel()
+		m := mapper.CreateSearchPage(cfg, req, basePage, validatedQueryParams, []data.Category{}, searchResp, departmentResp, lang, validationProblem)
```
nshumoogum, line 47: Do we need to pass in validateProblem into function? If we do then set this to true

I guess this can be used to determine an error message back in the template for the user to read?

nshumoogum (changes requested):

model/search.go
```diff
@@ -11,12 +11,13 @@ type SearchPage struct {
...
 // Search represents all search parameters and response data of the search
 type Search struct {
-	Query      string           `json:"query"`
-	Filter     []string         `json:"filter,omitempty"`
-	Filters    []Filter         `json:"filters"`
-	Sort       Sort             `json:"sort,omitempty"`
-	Pagination model.Pagination `json:"pagination,omitempty"`
-	Response   Response         `json:"response"`
+	Query             string           `json:"query"`
+	ValidationProblem string           `json:"ValidationProblem"`
```
nshumoogum, line 15: I would update this to be:

```suggestion
	ErrorMessage string           `json:"error_message,omitempty"`
```

json fields should be `camel_case` and the omitempty will hide this field in response if it is empty `""`

nshumoogum (changes requested):

nshumoogum (approved):

PR #103, opened 2022-04-01
add ENABLE_CENSUS_TOPIC_FILTER_OPTION flag

jondewijones:
### What

add `ENABLE_CENSUS_TOPIC_FILTER_OPTION` flag

### Who can review

Anyone

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1,27 +1,29 @@
 package config
 
 import (
-	"github.com/kelseyhightower/envconfig"
 	"time"
+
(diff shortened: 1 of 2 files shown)


rafahop (commented):
It looks good but can we amend the [unit test](https://github.com/ONSdigital/dp-frontend-search-controller/blob/14bd3f1c87335cbd8ef6fbb6c652a915b0713c73/config/config_test.go#L37) to check the flag is off by default?

rafahop (approved):

PR #113, opened 2022-04-19
Topic filter

liamjlewis:
### What

I have added the topics categories to work the same way as regular categories in anticipation of topics coming from the API.
https://trello.com/c/RX2xaXgE/1090-search-controller-update-controller-data-model-with-topics-filter
https://trello.com/c/upcLpqFt/1095-search-controller-add-checkbox-to-search-results-page

### How to review

In handlers/handlers.go line 153 change count to 1 to see the changes appear visually

Styling is found in the branch of dp design system
https://github.com/ONSdigital/dp-design-system/tree/topic-filter

### Who can review

Anyone.

### DIFF
--- a/assets/locales/service.cy.toml
+++ b/assets/locales/service.cy.toml
@@ -37,6 +37,51 @@ other = "Data"
 description = "Other"
 other = "Other"
 
+[DemographyAndMigration]
+description = "Demography and migration"
+one = "Demography and migration"
+other = "Demography and migration"
+
+[Education]
+description = "Education"
+one = "Education"
+other = "Education"
+
+[EthnicGroupNationalIdentityAndReligion]
+description = "Ethnic group, national identity and religion"
+one = "Ethnic group, national identity and religion"
+other = "Ethnic group, national identity and religion"
+
+[HealthDisabilityAndUnpaidCare]
+description = "Health, disability and unpaid care"
+one = "Health, disability and unpaid care"
+other = "Health, disability and unpaid care"
+
+[Housing]
+description = "Housing"
+one = "Housing"
+other = "Housing"
+
+[LabourMarketAndTravelToWork]
+description = "Labour market and travel to work"
+one = "Labour market and travel to work"
+other = "Labour market and travel to work"
+
+[SexualOrientationAndGenderIdentity]
+description = "Sexual orientation and gender identity"
+one = "Sexual orientation and gender identity"
+other = "Sexual orientation and gender identity"
+
+[Veteran]
+description = "Veteran"
+one = "Veteran"
+other = "Veterans"
+
+[Census]
+description = "Census"
+one = "Census"
+other = "Census"
+
 [StatisticalBulletin]
 description = "Statistical bulletins"
 one = "Bwletin ystadegol"
--- a/assets/templates/partials/filter.tmpl
+++ b/assets/templates/partials/filter.tmpl
@@ -1,8 +1,70 @@
 {{ $filters := .Data.Filters }}
+{{ $topicFilters := .Data.TopicFilters }}
 {{ $response := .Data.Response }}
 {{ $query := .Data.Query }}
 {{ $sort := .Data.Sort.Query }}
 {{ $lang := .Language }}
+
+<form id="topicsFilterForm" method="get" class="js-auto-submit__form" action="/search">
+    <input type="hidden" name="q" value="{{ $query }}"/>
+    <input type="hidden" name="sort" value="{{$sort}}"/>
+    <fieldset class="ons-fieldset">
+        {{ range $index, $topFilter := $topicFilters}}
+            {{if gt $topFilter.NumberOfResults 0}}
+                <div class="ons-checkboxes__items">
+                    <span class="ons-checkboxes__item ons-checkboxes__item--no-border">
+                        <span class="ons-checkbox ons-checkbox--no-border category-filter">
+                            <input type="checkbox" id="topic-group-{{ $index }}"
+                                class="ons-checkbox__input ons-js-checkbox"
+                                value="{{ print $topFilter.FilterKey }}" categoryChildren="{{ print $topFilter.FilterKey }}" aria-controls="topic-group-{{ $index }}-other-wrap"
+                                aria-haspopup="true"
+                                data-gtm-label="{{ $topFilter.LocaliseKeyName }}"
+                                {{if $topFilter.IsChecked }}
+                                    checked
+                                {{end}}
+                            >
+                            <label class="ons-checkbox__label" for="topic-group-{{ $index }}" id="topic-group-{{ $index }}-label">
+                                {{ localise $topFilter.LocaliseKeyName $lang 4 }} ({{ $topFilter.NumberOfResults }})
+                            </label>
+                            <span class="ons-checkbox__other" id="topic-group-{{ $index }}-other-wrap">
+                                <fieldset class="ons-fieldset ons-js-other-fieldset">
+                                    {{range $index, $childFilter := $topFilter.Types}}
+                                        {{if ne $childFilter.NumberOfResults 0}}
+                                            {{$id := index $childFilter.FilterKey 0}}
+                                            <div class="ons-checkboxes__items">
+                                                <span class="ons-checkboxes__item ons-checkboxes__item--no-border">
+                                                    <span class="ons-checkbox ons-checkbox--no-border child-filter">
+                                                        <input id="{{ $id }}"
+                                                            class="ons-checkbox__input ons-js-checkbox"
+                                                            type="checkbox" name="filter"
+                                                            {{ if $childFilter.IsChecked }}checked{{end}}
+                                                            data-gtm-label="{{ $childFilter.LocaliseKeyName }}"
+                                                            value="{{ $id }}"
+                                                        />
+                                                        <label class="ons-checkbox__label" for="{{ $id }}"
+                                                            id="{{ $id }}-label">
+                                                            {{ localise $childFilter.LocaliseKeyName $lang 4 }} ({{ $childFilter.NumberOfResults }})
+                                                        </label>
+                                                    </span>
+                                                </span>
+                                            </div>
+                                        {{ end }}
+                                    {{ end }}
+                                </fieldset>
+                            </span>
+                        </span>
+                    </span>
+                </div>
+            {{ end }}
+        {{ end }}
+    </fieldset>
+    <noscript>
+        <button type="submit"
+                class="btn btn--primary btn--thick margin-top-sm--2 margin-top-md--2 js-submit-button">Filter
+        </button>
+    </noscript>
+</form>
+
 <form id="filterForm" method="get" class="js-auto-submit__form" action="/search">
     <input type="hidden" name="q" value="{{ $query }}"/>
     <input type="hidden" name="sort" value="{{$sort}}"/>
--- a/assets/templates/partials/filter.tmpl
+++ b/assets/templates/partials/filter.tmpl
@@ -9,7 +9,7 @@
     <input type="hidden" name="q" value="{{ $query }}"/>
     <input type="hidden" name="sort" value="{{$sort}}"/>
     <fieldset class="ons-fieldset">
-        {{ range $index, $topFilter := $topicFilters}}
+        {{ range $index, $topFilter := $topicFilters }}
             {{if gt $topFilter.NumberOfResults 0}}
                 <div class="ons-checkboxes__items">
                     <span class="ons-checkboxes__item ons-checkboxes__item--no-border">
--- a/data/topic-filter.go
+++ b/data/topic-filter.go
@@ -1,0 +1,222 @@
+package data
+
+import (
+	"context"
+	"net/url"
+	"strings"
+
+	errs "github.com/ONSdigital/dp-frontend-search-controller/apperrors"
+	"github.com/ONSdigital/log.go/v2/log"
+)
+
+// Filter represents information of filters selected by user
+type TopicFilter struct {
+	Query           []string `json:"query,omitempty"`
+	LocaliseKeyName []string `json:"localise_key,omitempty"`
+}
+
+// TopicCategory represents all the search categories in search page
+type TopicCategory struct {
+	LocaliseKeyName string             `json:"localise_key"`
+	Count           int                `json:"count"`
+	ContentTypes    []TopicContentType `json:"content_types"`
+}
+
+// ContentType represents the type of the search results and the number of results for each type
+type TopicContentType struct {
+	LocaliseKeyName string   `json:"localise_key"`
+	Count           int      `json:"count"`
+	Group           string   `json:"group"`
+	Types           []string `json:"types"`
+	ShowInWebUI     bool     `json:"show_in_web_ui"`
+}
+
+var defaultTopicContentTypes = "article," +
+	"article_download," +
+	"bulletin," +
+	"compendium_landing_page," +
+	"dataset_landing_page," +
+	"product_page," +
+	"static_adhoc," +
+	"static_article," +
+	"static_foi," +
+	"static_landing_page," +
+	"static_methodology," +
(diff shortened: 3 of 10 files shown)


assets/templates/partials/filter.tmpl
```diff
@@ -1,8 +1,70 @@
...
 {{ $response := .Data.Response }}
 {{ $query := .Data.Query }}
 {{ $sort := .Data.Sort.Query }}
 {{ $lang := .Language }}
+
+<form id="topicsFilterForm" method="get" class="js-auto-submit__form" action="/search">
+    <input type="hidden" name="q" value="{{ $query }}"/>
+    <input type="hidden" name="sort" value="{{$sort}}"/>
+    <fieldset class="ons-fieldset">
+        {{ range $index, $topFilter := $topicFilters}}
```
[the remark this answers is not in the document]
liamjlewis: haha yeah that is pretty frustrating to look at

assets/locales/service.cy.toml
```diff
@@ -37,6 +37,51 @@ other = "Data"
 description = "Other"
 other = "Other"
 
+[DemographyAndMigration]
```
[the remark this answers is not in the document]
liamjlewis: Yeah, I wasn't given any. This whole topic thing is going to be a work in progress for a while I think because the API doesn't yet provide topics.

data/topic-filter.go
```diff
@@ -0,0 +1,222 @@
+package data
```
jondewijones, line 1: is this a c+p of filter.go? do we need all of it? can we re-use parts from both of these to reduce duplication? can see parts here we don't need e.g. `defaultTopicContentTypes` since those are used for the `?filter=` filter not topic filter

mapper/mapper.go
```diff
@@ -321,6 +323,42 @@ func mapFilters(page *model.SearchPage, categories []data.Category, queryParams
 	page.Data.Filters = filters
 }
 
+func mapTopicFilters(page *model.SearchPage, topicCategories []data.TopicCategory, queryParams data.SearchURLParams) {
```
jondewijones, line 326: similar as previous comment; do we need to duplicate this functionality when it's doing the same/virtually the same?

data/topic-filter.go
```diff
@@ -0,0 +1,222 @@
...
+		Housing.Group:                                Housing,
+		LabourMarketAndTravelToWork.Group:            LabourMarketAndTravelToWork,
+		SexualOrientationAndGenderIdentity.Group:     SexualOrientationAndGenderIdentity,
+		Veterans.Group:                               Veterans,
+	}
+)
+
+// reviewFilter retrieves filters from query, checks if they are one of the filter options, and updates validatedQueryParams
+func reviewTopicFilters(ctx context.Context, urlQuery url.Values, validatedQueryParams *SearchURLParams) error {
+	topicFiltersQuery := urlQuery["topic"]
```
nshumoogum, line 140: Shouldn't the value of query parameter be `topics` instead of `topic`? We can have more than one comma separated list of topics

nshumoogum: So we don't want the same feature in the url as filter.

The url should look like: `ons.gov.uk/search?q=xxx&topics=census-age,census-health-and-wellbeing` **not** `ons.gov.uk/search?q=xxx&topic=census-age&topic=census-health-and-wellbeing`

Therefore the `topicFiltersQuery` is of type string and probably renamed to just `topicFilters`

nshumoogum: As it is a string, you will then need to split the string by comma into a new type []string

like so:

```
    topicFilters := urlQuery["topics"]
    topics := strings.Split(topicFilters, ",")

    for _, topic := topics {
        ...
```

data/topic-filter.go
```diff
@@ -0,0 +1,222 @@
...
+	"context"
+	"net/url"
+	"strings"
+
+	errs "github.com/ONSdigital/dp-frontend-search-controller/apperrors"
+	"github.com/ONSdigital/log.go/v2/log"
+)
+
+// Filter represents information of filters selected by user
+type TopicFilter struct {
```
nshumoogum, line 13: TopicFilter object is the same as Filter struct - so no need for TopicFilter struct just use Filter struct (use inheritance). The Filter struct does have a generic name so it shouldn't be confusing

data/topic-filter.go
```diff
@@ -0,0 +1,222 @@
...
+	Query           []string `json:"query,omitempty"`
+	LocaliseKeyName []string `json:"localise_key,omitempty"`
+}
+
+// TopicCategory represents all the search categories in search page
+type TopicCategory struct {
+	LocaliseKeyName string             `json:"localise_key"`
+	Count           int                `json:"count"`
+	ContentTypes    []TopicContentType `json:"content_types"`
```
nshumoogum, line 22: This surely should be `Topics` instead of `ContentTypes` and should be of type `[]Topic` and json tag should be `topics`?

```suggestion
	Topics    []Topic `json:"topics"`
```

data/topic-filter.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+// ContentType represents the type of the search results and the number of results for each type
+type TopicContentType struct {
```
nshumoogum, line 26: Rename struct to `Topic` as per comment above?

data/topic-filter.go
```diff
@@ -0,0 +1,222 @@
...
+	LocaliseKeyName string   `json:"localise_key"`
+	Count           int      `json:"count"`
+	Group           string   `json:"group"`
+	Types           []string `json:"types"`
```
nshumoogum, line 30: Is this field just the subtopics? So would it not be better to rename it so?

```suggestion
	SubTopics           []string `json:"subtopics"`
```

data/topic-filter.go
```diff
@@ -0,0 +1,222 @@
...
+	ShowInWebUI     bool     `json:"show_in_web_ui"`
+}
+
+var defaultTopicContentTypes = "article," +
```
nshumoogum, line 34: These are the default topics not content types, should we not be using a different list?

If we don't know the values then leave this as empty with a TODO comment to fill in a list of default topics

nshumoogum:
```suggestion
// TODO extend default topics with list of topics
var defaultTopics = ""
```

nshumoogum: So just seen below we do have the list of topics; the `census` topic

data/topic-filter.go
```diff
@@ -0,0 +1,222 @@
...
+		for j := range topicCategory.ContentTypes {
+			topicCategories[i].ContentTypes[j].Types = []string{}
+			topicCategories[i].ContentTypes[j].Types = append(topicCategories[i].ContentTypes[j].Types, TopicCategories[i].ContentTypes[j].Types...)
+		}
+	}
+
+	return topicCategories
+}
+
+// updateQueryWithAPITopicFilters retrieves and adds all available sub filters which is related to the search filter given by the user
```
nshumoogum, line 187:
```suggestion
// updateQueryWithAPITopicFilters adds all available subtopic filters to the API query
```

data/topic-filter.go
```diff
@@ -0,0 +1,222 @@
...
+func updateQueryWithAPITopicFilters(apiQuery url.Values) {
+	filters := apiQuery["content_type"]
```
nshumoogum, line 189: This should be `topics` instead of `content_type`

data/topic-filter.go
```diff
@@ -0,0 +1,222 @@
...
+		subFilters := getTopicSubFilters(filters)
+
+		apiQuery.Set("content_type", strings.Join(subFilters, ","))
+	} else {
+		apiQuery.Set("content_type", defaultTopicContentTypes)
+	}
+}
+
+// getTopicSubFilters gets all available sub filters which is related to the search filter given by the user
+func getTopicSubFilters(filters []string) []string {
```
nshumoogum, line 201: A better name for this function might be `getListOfSubTopics` and then comment updated in suggestion below too

```
// getListOfSubTopics gets all available subtopics for topics which have been set to filter results by  the user
func getTopicSubFilters(filters []string) []string {
```

data/topic-filter.go
```diff
@@ -0,0 +1,222 @@
...
+	var subFilters = make([]string, 0)
```
nshumoogum, line 202: Probably worth changing the variable name to subTopics

```suggestion
	var subTopics []string
```

data/topic-filter.go
```diff
@@ -0,0 +1,222 @@
...
+
+	for _, filter := range filters {
+		subFilter := topicFilterOptions[filter]
```
nshumoogum, line 205: Again `subTopic` instead?

model/search.go
```diff
@@ -29,6 +30,15 @@ type Filter struct {
 	Types           []Filter `json:"types,omitempty"`
 }
 
+// TopicFilter respresents all topic filter information needed by templates
+type TopicFilter struct {
```
nshumoogum, line 34: I don't think you need this as it is a duplicate of Filter struct

nshumoogum: Think about updating `Types` to `Filters` to make it more generic on line 39

model/search.go
```diff
@@ -15,6 +15,7 @@ type Search struct {
 	ErrorMessage string           `json:"error_message,omitempty"`
 	Filter       []string         `json:"filter,omitempty"`
 	Filters      []Filter         `json:"filters"`
+	TopicFilters []TopicFilter    `json:"topic_filters"`
```
nshumoogum, line 18: Can just use Filter struct

```suggestion
	TopicFilters []Filter    `json:"topic_filters"`
```

data/topic-filter.go
nshumoogum: So I've added comment addressing renaming things related to topics and removing unnecessary duplication. We could go a step further and most of these functions could be turned into ones that handle interfaces instead of TopicFilter and ContentTypeFilter (which is described Filter everywhere). That way there would be even less duplication and easier to extend later down the line when another filter option becomes available.

I suggest this can be looked at in a separate ticket.

nshumoogum (changes requested):

data/topic-filter.go
```diff
@@ -0,0 +1,213 @@
...
+		Housing.Group:                                Housing,
+		LabourMarketAndTravelToWork.Group:            LabourMarketAndTravelToWork,
+		SexualOrientationAndGenderIdentity.Group:     SexualOrientationAndGenderIdentity,
+		Veterans.Group:                               Veterans,
+	}
+)
+
+// reviewFilter retrieves filters from query, checks if they are one of the filter options, and updates validatedQueryParams
+func reviewTopicFilters(ctx context.Context, urlQuery url.Values, validatedQueryParams *SearchURLParams) error {
+	topicFilters := urlQuery["topics"]
```
nshumoogum, line 125: So if you use `urlQuery.Get("topics")` this will get the first case of topics in path

e.g. /something?topics="health,welfare"&topics="ethnicity"
The Get method will retrieve `"health,welfare"` only as `string` instead of `[]string`

This is better for us as we wont need lines 127 to 129

nshumoogum: More like this:

```
func reviewTopicFilters(ctx context.Context, urlQuery url.Values, validatedQueryParams *SearchURLParams) error {
	topicFilters := urlQuery.Get("topics")

	topics := strings.Split(topicFilters, ",")

	for _, topicFilterQuery := range topics {
        ...
}
```

data/topic-filter.go
nshumoogum: Suggest changing this comment to the one I suggested above - clean and concise; no need to mention that a user inputs these filters

```suggestion
// updateQueryWithAPITopicFilters adds all available subtopic filters to the API query
```

data/topic-filter.go
```diff
@@ -0,0 +1,213 @@
...
+	} else {
+		apiQuery.Set("topics", defaultTopics)
+	}
+}
+
+// getListOfSubTopics gets all available subtopics for topics which have been set to filter results by the user
+func getListOfSubTopics(topics []string) []string {
+	var subTopics []string
+
+	for _, topic := range topics {
```
nshumoogum, line 195: Suggest we make this more performant by not creating topic variables and therefore increasing the memory footprint. We do this by changing the code to do the following:

```suggestion
	for i := range topics {
```

nshumoogum: Line 196 will then need to change to:

```
    topicFilterOption := topicFilterOptions[topics[i]]
```

Also in the above noticed I've changed the variable name, this aligns with the naming convention of topicFilter

data/topic-filter.go
```diff
@@ -0,0 +1,213 @@
...
+		subFilter := topicFilterOptions[topic]
+		subTopics = append(subTopics, subFilter.SubTopics...)
```
nshumoogum, line 197: As mentioned above, we will want to change the `subFilter` variable name to `topicFilterOption`:

```suggestion
		subTopics = append(subTopics, topicFilterOption.SubTopics...)
```

data/topic-filter.go
```diff
@@ -0,0 +1,213 @@
...
+	return nil
+}
+
+// GetTopicCategories returns all the categories and its content types where all the count is set to zero
+func GetTopicCategories() []TopicCategory {
+	var topicCategories []TopicCategory
+	topicCategories = append(topicCategories, TopicCategories...)
+
+	// To get a different reference of Topic - deep copy
+	for i, topicCategory := range topicCategories {
```
nshumoogum, line 164: Suggest change this to:

```suggestion
	for i := range topicCategories {
```

And then anywhere that uses `topicCategory` change to , `topicCategories[i]`

data/topic-filter.go
```diff
@@ -0,0 +1,213 @@
...
+		subFilter := topicFilterOptions[topic]
+		subTopics = append(subTopics, subFilter.SubTopics...)
+	}
+
+	return subTopics
+}
+
+// GetTopicGroupLocaliseKey gets the localise key of the group type of the search result to be displayed
+func GetTopicGroupLocaliseKey(resultType string) string {
+	for _, filterOption := range topicFilterOptions {
```
nshumoogum, line 205: Suggest changing this so it does not create multiple copies of topicFilterOptions[i] in memory, if we could change this to do the following:

```suggestion
	for i := range topicFilterOptions {
```

Wherever we previously used `filterOption` replace with `topicFilterOptions[i]`

data/topic-filter.go
```diff
@@ -0,0 +1,213 @@
...
+		for _, optionType := range filterOption.SubTopics {
```
nshumoogum, line 206: As above, suggest changing this to:

```suggestion
		for j := range topicFilterOptions[i].SubTopics {
```

And wherever `optionType` was being used change to `topicFilterOptions[i].SubTopics[j]`

data/topic-filter.go
```diff
@@ -0,0 +1,213 @@
...
+func reviewTopicFilters(ctx context.Context, urlQuery url.Values, validatedQueryParams *SearchURLParams) error {
+	topicFilters := urlQuery["topics"]
+
+	if topicFilters == nil {
+		return nil
+	}
+
+	topics := strings.Split(topicFilters[0], ",")
+
+	for _, topicFilterQuery := range topics {
```
nshumoogum, line 133: Another for loop which could do with memory usage improvement, suggest doing the following:

```suggestion
	for i := range topics {
```

Where `topicFilterQuery` is being used in function to change this to `topics[i]`

data/topic-filter.go
```diff
@@ -0,0 +1,213 @@
...
+
+		topicFilterQuery = strings.ToLower(topicFilterQuery)
+
+		if topicFilterQuery == "" {
+			continue
+		}
+
+		topicFilter, found := topicFilterOptions[topicFilterQuery]
+
```
nshumoogum, line 142: no need for white space

nshumoogum (changes requested):

nshumoogum (approved):

PR #115, opened 2022-04-19
translation for the filter

liamjlewis:
### What

Added some translation for the filter items and also the heading "Content type"

### How to review

Read the code or run locally if you like.

### Who can review

Anyone

### DIFF
--- a/assets/locales/service.cy.toml
+++ b/assets/locales/service.cy.toml
@@ -23,24 +23,24 @@ other = "Mathau o gynnwys i'w dangos"
 
 [ContentType]
 description = "Content type"
-one = "math o gynnwys"
-other = "math o gynnwys"
+one = "Math o gynnwys"
+other = "Math o gynnwys"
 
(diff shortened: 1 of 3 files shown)


assets/locales/service.cy.toml
```diff
@@ -21,21 +21,26 @@ other = "Mireinio'r canlyniadau"
 description = "Content types to show"
 other = "Mathau o gynnwys i'w dangos"
 
+[ContentType]
+description = "Content type"
+one = "math o gynnwys"
```
gilesburdett, line 26: I suspect the first letter of a number of these entries should be capitalised in Welsh as they are in English

gilesburdett (changes requested):

assets/locales/service.cy.toml
liamjlewis: yes indeed! I have corrected.

gilesburdett (approved):

PR #117, opened 2022-04-27
a loading message on the search page

liamjlewis:
### What

A loading indicator to show results are taking a while to arrive, there is a data label included as an alternative message in the case of server issues.

> As a website user, with a slow internet connection
> I want to know that my search worked
> so that I am reassured and do not refresh
> 
> 

Required dp-design-system changes: https://github.com/ONSdigital/dp-design-system/pull/85

https://trello.com/c/08BLV9Xr/1058-add-loading-indicator-to-search-results-page?filter=label:Front%20End

### How to review

Read or pull. Don't forget the dp-design-system counterpart listed above.

### Who can review

Anyone.

### DIFF
--- a/assets/templates/partials/results.tmpl
+++ b/assets/templates/partials/results.tmpl
@@ -4,6 +4,14 @@
 {{ $totalSearchPosition := multiply (subtract $currentPage 1) $itemsPerPage }}
 {{ $response := .Data.Response }}
 <div id="results">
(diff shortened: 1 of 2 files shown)


mr-nick17 (commented):
No problems with the code here but I have suggested some changes to your PR in dp-design-system so will need to be re-reviewed

gilesburdett (approved):

PR #119, opened 2022-05-16
Feature/emergency and service banners

NoelleDL:
### What

Make emergency and service banner appear on new search.; add health check for Zebedee. 

### How to review

Must have homepage content locally with a service message and emergency banner information. Banner and service message are currently not using the new design system so banner will appear white. 

### Who can review

Anyone


service/service.go
```diff
@@ -49,11 +51,16 @@ func (svc *Service) Init(ctx context.Context, cfg *config.Config, serviceList *E
...
 	clients := routes.Clients{
-		Search: search.NewWithHealthClient(svc.routerHealthClient),
+		Renderer: render.NewWithDefaultClient(assets.Asset, assets.AssetNames, cfg.PatternLibraryAssetsPath, cfg.SiteDomain),
+		Search:   search.NewWithHealthClient(svc.routerHealthClient),
+		Zebedee:  zebedee.NewWithHealthClient(svc.routerHealthClient),
 	}
 
 	// Initialise render client, routes and initialise localisations bundles
-	rend := render.NewWithDefaultClient(assets.Asset, assets.AssetNames, cfg.PatternLibraryAssetsPath, cfg.SiteDomain)
+	//Rend := render.NewWithDefaultClient(assets.Asset, assets.AssetNames, cfg.PatternLibraryAssetsPath, cfg.SiteDomain)
```
justinpjose, line 60: These comments can be removed

features/steps/component.go
```diff
@@ -87,11 +90,17 @@ func (c *Component) InitAPIFeature() *componentTest.APIFeature {
 	return c.APIFeature
 }
 
-// Reset resets the search controller component
-func (c *Component) Reset() *Component {
+// Reset resets the search controller component to its default values
+func (c *Component) Reset() error {
+	if c.cfg == nil {
```
nshumoogum, line 95: So not sure the Reset method is being used as a reset?

It is being used to set the configuration if it is nil, that means nothing is being reset.

nshumoogum: Think about creating a new method if you want this functionality,`func SetConfig() error`

Im still unsure (at this moment in peer review) why you need this functionality

mapper/mapper.go
```diff
@@ -12,7 +15,8 @@ import (
 )
 
 // CreateSearchPage maps type searchC.Response to model.Page
-func CreateSearchPage(cfg *config.Config, req *http.Request, basePage coreModel.Page, validatedQueryParams data.SearchURLParams, categories []data.Category, topicCategories []data.TopicCategory, respC searchC.Response, departments searchC.Department, lang string, ErrorMessage string) model.SearchPage {
+
+func CreateSearchPage(cfg *config.Config, req *http.Request, basePage coreModel.Page, validatedQueryParams data.SearchURLParams, categories []data.Category, topicCategories []data.TopicCategory, respC searchC.Response, departments searchC.Department, lang string, homepageResponse zebedee.HomepageContent, ErrorMessage string) model.SearchPage {
```
nshumoogum, line 19: It might be worth spreading this very long function signature across at least 2 lines

mapper/mapper_test.go
```diff
@@ -49,14 +47,18 @@ func TestUnitCreateSearchPageSuccess(t *testing.T) {
...
+
 		respC, err := GetMockSearchResponse()
 		So(err, ShouldBeNil)
 
 		respD, err := GetMockDepartmentResponse()
 		So(err, ShouldBeNil)
 
 		Convey("When CreateSearchPage is called", func() {
-			sp := CreateSearchPage(cfg, req, mdl, validatedQueryParams, categories, topicCategories, respC, respD, lang, "")
+
```
nshumoogum, line 60: Consider removing whitespace

nshumoogum (changes requested):

nshumoogum:
It might be worth us going through together the component test setup; just to tidy up how the component and external dependencies are setup

justinpjose:
> It might be worth us going through together the component test setup; just to tidy up how the component and external dependencies are setup

With regards to the component test, a few things were done of the following:

**Removing `Reset` and `BeforeScenario` function**
- For each scenario in the test, a new controller component is created with its respective APIFeature. A fakeAPIRouter is also created within the controller component
- The newly created controller component, APIFeature, and fakeAPIRouter are reset (empty) by default when they are created. Therefore, there is no need to call a `Reset` function for these features in the `BeforeScenario` function. Therefore, the `Reset` and `BeforeScenario` were removed
- When the Scenario ends, it is important to close the controller component and fakeAPIRouter (essentially closing their server) otherwise these will be still running in the background. This is why the `AfterScenario` function is important

**Increasing the `HealthCheckCriticalTimeout`**
- Whilst running the component tests 30 times consecutively, I noticed that there were issues with timings. At the time, within the component test, the `HealthCheckInterval` was set to `1s` and the `HealthCheckCriticalTimeout` was set to `2s`.  - In the scenario `Returning a WARNING (429) status when one downstream service is critical and critical timeout has not expired`, even though the test passes with this configuration most of the time, in some cases, the test failed as the test returned a `500` status code. 
- This is because in that particular case, the `/health` was called by the service after the `HealthCheckCriticalTimeout` even though we have specifically told it to call before. This was a strange one to solve but I then realized that the `HealthCheckInterval` and `HealthCheckCriticalTimeout` were close to each other (1s difference) in which the performance of the component running can allow for a change in result, especially with a close time range (1s difference). 
- As a result, I increased the `HealthCheckCriticalTimeout` to `3s` which gave more room to avoid this issue, and ran the component test 30 times to which this passed.  

**Removing check for `LastSuccess` for healthcheck fails and `LastFailure` for healthcheck passes**
- Whilst running the component tests 30 times consecutively, we also experienced some test failures with the error `invalid memory address or nil pointer dereference` on some infrequent occasions. To be exact, out of the 4 scenarios which were run 30 times (120 scenarios run in total), this error only occurred in 2 out of the 120. 
- This specifically occurred in scenarios where the health does not return a 200. 
- When investigating those specific occasions, the pointer `LastSuccess` was `nil` when the healthcheck fails, even though we were expecting a specific date-time at this instance
- For context, when the service runs, it calls `/health` to the downstream apps (just `API router` in our case) and it does this repeatedly on each `HealthCheckInterval`. We have set the component test to return 200 for this when the service begins initially and then this is changed accordingly for each scenario. Most of the time, some `/health` calls are made to the `fakeAPIRouter` before the scenario gets to run (this happens because the service is run and then the scenarios are called). During that time, the `LastSuccess` is populated with a date-time. 
- But as discovered, in some rare instances, the `LastSuccess` was `nil` and this occurred because, in that instance, the first `/health` call was made during the scenario rather than before the scenario is run. To be more specific, as we tell the `fakeAPIRouter` to return a non-200 in the scenario - when the first `/health` happens during the scenario, the `LastSuccess` cannot be populated as there was no previous success health in the first place. 
- To overcome this, we can add a delay after the service runs in the component to ensure that `/health` calls are made before the scenario is run. But this slows the component test even more for a check which isn't significant. 
- Therefore, the check of `LastSuccess` in failed healthcheck scenarios was removed. For consistency, the check of `LastFailure` was also removed for 200 health scenario. This means that we just check if the `LastFailure` is populated for a failure and `LastSuccess` is populated for a success - straightforward and simple

**Adding `mutex.Lock` and `Unlock` for `healthRequest`**
- When running the component test, there was a data race issue. This data race occurred because we were writing and reading the `healthRequest` of the `fakeAPIRouter` at the same time. To be more specific, we were writing when changing the status code return of the `healthRequest` (`/health`) of the `fakeAPIRouter` for each scenario in the test. We were reading when the (`/health`) calls were made from the service (at each `HealthCheckInterval`). 
- To overcome this, `mutex.Lock` and `mutex.Unlock` were added when changes were made to the `healthRequest` (`/health`) of the `fakeAPIRouter` which resolved this issue

I can confirm that I re-ran the component tests 30 times consecutively and no errors came up as the issues were resolved as mentioned above

features/steps/component.go
justinpjose: I have addressed this in the big comment below

nshumoogum (approved):
Component tests will be looked at in separate PR

PR #134, opened 2022-06-17
Configure search no index enabled

cookel2:
# What has changed

1. The latest changes have been pulled in from dp-renderer.

2. The README has been updated to include the NO_INDEX_ENABLED variable. It has also been tidied up so that the correct variables are listed and are alphabetical (matching config.go).

3. The SearchNoIndexEnabled Page value, which is defined in the dp-renderer, gets set to the value of NO_INDEX_ENABLED.

# How to review

Check that the changes make sense and that the tests pass:

make test

# Who can review

Anyone but me


README.md
```diff
@@ -20,24 +20,25 @@ No further dependencies other than those defined in `go.mod`
...
+| DEFAULT_MAXIMUM_SEARCH_RESULTS    | 500                          | The default maximum search results
+| DEFAULT_OFFSET                    | 0                            | The default offset of search results
+| DEFAULT_PAGE                      | 1                            | The default current page of search results
+| DEFAULT_SORT                      | relevance                    | The default sort of search results
+| ENABLE_CENSUS_TOPIC_FILTER_OPTION | false                        | Enable filtering on various census topics
+| GRACEFUL_SHUTDOWN_TIMEOUT         | 5s                           | The graceful shutdown timeout in seconds (`time.Duration` format)
+| HEALTHCHECK_CRITICAL_TIMEOUT      | 90s                          | Time to wait until an unhealthy dependent propagates its state to make this app unhealthy (`time.Duration` format)
+| HEALTHCHECK_INTERVAL              | 30s                          | Time between self-healthchecks (`time.Duration` format)
+| NO_INDEX_ENABLED                  | false                        | If true then prevents the ONS search results from appearing in Google Search
```
justinpjose, line 38: In our case, it is most search engine web crawlers and not just Google

```suggestion
| NO_INDEX_ENABLED                  | false                        | If true then this prevents most search engine web crawlers from indexing the search pages
```

justinpjose (changes requested):

justinpjose (commented):
I would also update the `mapper_test.go` file to include `NO_INDEX_ENABLED` as well

PR #138, opened 2022-06-24
cache census topics in search controller

justinpjose:
### What
- Cache census topic data
- Update cache every 30mins from startup

### How to review
- Check if the test passes

### Who can review
- Anyone


cache/census_topic_pub.go
```diff
@@ -0,0 +1,131 @@
...
+func getSubtopicsIDsPrivate(ctx context.Context, subtopicsIDChan chan string, topicClient topicCli.Clienter, topLevelTopicID string) {
+	// get subtopics from dp-topic-api
+	subTopics, err := topicClient.GetSubtopicsPrivate(ctx, topicCli.Headers{}, topLevelTopicID)
+	if err != nil {
+		if err != topicCliErr.ErrNotFound {
+			logData := log.Data{
+				"req_headers":        topicCli.Headers{},
+				"top_level_topic_id": topLevelTopicID,
+			}
+			log.Error(ctx, "failed to get subtopics from topic-api", err, logData)
```
rahulPalliyalil, line 99: (minor): you would probably need to move the ```log.Error``` message  outside of ``` err != topicCliErr.ErrNotFound``` loop as this should log regardless of the error type right ? I mean you would have to log for any err != nil ?

cache/census_topic_pub.go
```diff
@@ -0,0 +1,131 @@
...
+		// stop as there are no subtopics items or failed to get subtopics
+		return
+	}
+
+	//deference sub topics items to allow ranging through them
+	var subTopicItems []models.TopicResponse
+	if subTopics.PrivateItems != nil {
+		subTopicItems = *subTopics.PrivateItems
+	} else {
+		err := errors.New("sub topics public items is nil")
```
rahulPalliyalil, line 111: (question): Do you mean private items  ?  ```sub topics public items is nil``` , you are putting this on the check ```subTopics.PrivateItems == nil```` not public items ?

cache/census_topic_pub.go
```diff
@@ -0,0 +1,131 @@
...
+		log.Error(ctx, "failed to deference sub topics items pointer", err)
+		return
```
rahulPalliyalil, line 113: (suggestion): Also you could refactor this rather than if else, you can probably do: 

```
if subTopics.PrivateItems == nil {
  err := errors.New(".......")
  log.Error(ctx, "....", err)
  return
}
subTopicItems = *subTopics.PrivateItems
```
like this you don't need the else ? just a suggestion though

Makefile
```diff
@@ -31,7 +31,7 @@ lint:
 
 .PHONY: test
 test: generate-prod
-	go test -race -cover -tags 'production' ./...
+	go test -race -cover -count=1 -tags 'production' ./...
```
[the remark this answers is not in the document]
justinpjose: Good spot. I did it to run the test multiple times locally without caching. Will remove it

cache/census_topic_pub.go
justinpjose: In this case, it is expected and okay if no subtopics are found as this is what determines what the function to stop. If an error occurs for a different reason, then we need to have a look.

cache/census_topic_pub.go
```diff
@@ -0,0 +1,131 @@
...
+
+	topicCliErr "github.com/ONSdigital/dp-topic-api/apierrors"
+	"github.com/ONSdigital/dp-topic-api/models"
+	topicCli "github.com/ONSdigital/dp-topic-api/sdk"
+	"github.com/ONSdigital/log.go/v2/log"
+)
+
+// the model returned from the dp-topic-api is PrivateSubtopics in publishing mode (private)
+
+func UpdateCensusTopicPrivate(ctx context.Context, topicClient topicCli.Clienter) func() (interface{}, error) {
```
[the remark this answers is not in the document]
justinpjose: Definitely agreed and will do this in a future iteration

cache/census_topic_pub.go
justinpjose: Done

cache/census_topic_pub.go
justinpjose: Love this and Done

rahulPalliyalil (approved):

PR #139, opened 2022-06-25
add service auth token

justinpjose:
### What
- Add SERVICE_AUTH_TOKEN which is required to identify the controller when it calls the topic API via the API router in publishing mode
- This fixes issues for caching census topics in the controller when in publishing mode

### How to review

### Who can review
- Anyone

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,25 +20,28 @@ No further dependencies other than those defined in `go.mod`
 
 ## Configuration
 
-| Environment variable              | Default                      | Description
-| --------------------------------- | ---------------------------- | --------------------------------------------------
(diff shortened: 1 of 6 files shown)


cache/subtopic.go
```diff
@@ -0,0 +1,55 @@
...
+	return &SubtopicsIDs{
+		subtopicsMap: make(map[string]bool),
+	}
+}
+
+// Get returns a bool value for the given key (id) to inform that the subtopic id exists
+func (t *SubtopicsIDs) Get(key string) bool {
+	t.Lock()
+	defer t.Unlock()
+
```
rahulPalliyalil, line 27: (minor): I think you can allow multiple reads here as it is thread safe, so you could use a ```RWMutex``` in stead of ```Mutex``` here. With ```R.Lock() and R.unlock``` ?

rahulPalliyalil (approved):

PR #156, opened 2022-07-12
update census topic id

justinpjose:
### What
- `CensusTopicID` is unique in `sandbox` and `prod` so this should be configurable for environments
- Migrate to using `dp-cache` library

### How to review
- Check if the test passes

### Who can review
- Anyone


cache/topic.go
```diff
@@ -6,13 +6,13 @@ import (
...
-	"github.com/ONSdigital/dp-frontend-search-controller/dpcache"
+	dpcache "github.com/ONSdigital/dp-cache"
 	"github.com/ONSdigital/log.go/v2/log"
 )
 
-const (
+var (
 	// CensusTopicTitle is the id of the Census topic stored in mongodb which is accessible by using dp-topic-api
-	CensusTopicID = "4445"
+	CensusTopicID string
```
cookel2, line 15: In line 14 it needs to say CensusTopicID instead of CensusTopicTitle now.

cookel2 (changes requested):
Just one small comment.

PR #164, opened 2022-08-16
Nav from topic api

liamjlewis:
### What

Navigation bar data is taken from dp-topic-api and sent through to dp-renderer. It is cached using the dp-cache library.

### How to review

No special instructions.

### Who can review

Anyone.


justinpjose (approved):

cache/public/navigation.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/ONSdigital/log.go/v2/log"
+)
+
+func UpdateNavigationData(ctx context.Context, cfg *config.Config, lang string, topicClient topicCli.Clienter) func() *topicModel.Navigation {
+	if cfg.EnableNewNavBar == false {
+		return func() *topicModel.Navigation {
+			return &topicModel.Navigation{}
+		}
+	}
+	return func() *topicModel.Navigation {
```
justinpjose, line 18: Why are we returning a `func() *topicModel.Navigation`. Can we just not return `*topicModel.Navigation`?

cache/public/navigation_test.go
```diff
@@ -0,0 +1,91 @@
...
+		{
+			Description: "This is an item description.",
+			Label:       "Item label",
+			Links:       testTopicLinks,
+			Name:        "Item name.",
+			Title:       "Item title.",
+			Uri:         "Item URI.",
+		},
+	}
+	cfg, _ = config.Get()
```
justinpjose, line 49: You want to be calling this in each Test Function. And also, you don't want to omit the error even if it is used for testing purposes

Below `ctx := context.Background()` (line 56) in the func `TestUpdateNavigationData`, you should do

cfg, err := config.Get()
if err != nil {
     t.Errorf("failed to get config")
}

This will give the error and then fail the test

You should do this for each test function. If you are sharing configuration each test, it can led to a data race issue

cache/public/navigation.go
liamjlewis: This is what ultimately gets handed to the dp-cache update function, so I think it needs to be structured this way. Perhaps there's a more efficient way?

justinpjose: This makes sense. Also, looked at where this function is used and it all links up

justinpjose (approved):

PR #168, opened 2022-10-05
Remove logic to switch between ES 7.10 API response and ES 2.2 API response

nshumoogum:
### What

See title, but also includes ripping out departments from controller

### How to review

- check changes make sense
- check unit tests pass
- run search services locally to see controller responding correctly, services needed:
  - dp-search-api (I actually tunnelled into sandbox environment and used port forwarding)
  - dp-api-router
  - dp-topics-api
  - zebedee (homepage content needed, otherwise page will return 500)

### Who can review

Anyone


cache/caches.go
```diff
@@ -1,7 +1,7 @@
 package cache
 
 // CacheList is a list of caches for the dp-frontend-search-controller
-type CacheList struct {
+type List struct {
```
justinpjose, line 4: Any particular reason why this has been changed to `List`? Isn't CacheList more explicit about what it is?
Also, update needed for the comment to the new name change

ci/build.yml
```diff
@@ -6,7 +6,7 @@ image_resource:
   type: docker-image
   source:
     repository: golang
-    tag: 1.18.3
+    tag: 1.18.6
```
justinpjose, line 9: **COMMENT 1**

```suggestion
    tag: 1.19.1
```
Go version 1.19 is available

ci/lint.yml
```diff
@@ -6,7 +6,7 @@ image_resource:
   type: docker-image
   source:
     repository: golang
-    tag: latest
+    tag: 1.18.6
```
justinpjose, line 9: Same as **COMMENT 1**

ci/unit.yml
```diff
@@ -6,7 +6,7 @@ image_resource:
   type: docker-image
   source:
     repository: golang
-    tag: 1.18.3
+    tag: 1.18.6
```
justinpjose, line 9: Same as **COMMENT 1**

justinpjose (commented):
Some comments added

justinpjose (approved):

PR #169, opened 2022-10-07
Handle emergency and service banner gracefully

nshumoogum:
### What

See title.

If the app fails to get the banner data from Zebedee still load the page without the banners. Also includes:

- handle second request to search API as a parallel process to the first request
- if error is returned validating query parameters still load the page with no results and include banners if content is returned from Zebedee

### How to review

- Check changes makes sense
- Unit and component tests pass
- Linter and audit pass
- Emergency and service banner to appear if the user inputted less than three characters in search query term

### Who can review

Anyone


handlers/handlers.go
```diff
@@ -59,69 +59,85 @@ func read(w http.ResponseWriter, req *http.Request, cfg *config.Config, zc Zebed
 		return
 	}
 
-	apiQuery := data.GetSearchAPIQuery(validatedQueryParams, censusTopicCache)
-
-	var homepageResponse zebedeeCli.HomepageContent
-	var searchResp searchCli.Response
-	var respErr error
+	var (
+		counter            = 3
```
rafahop, line 63: I see this is for the wait group but I would add some quick comments explaining what this `counter` variable is for and why we are reducing it by 2 in line 71. It would make it a lot easier to understand if we don't have magic numbers.

handlers/handlers_test.go
```diff
@@ -300,7 +300,7 @@ func TestUnitReadFailure(t *testing.T) {
 				So(w.Code, ShouldEqual, http.StatusBadRequest)
 
 				So(len(mockedRendererClient.BuildPageCalls()), ShouldEqual, 0)
-				So(len(mockedSearchClient.GetSearchCalls()), ShouldEqual, 1)
+				So(len(mockedSearchClient.GetSearchCalls()), ShouldEqual, 2)
```
rafahop, line 303: It took me a while to understand why this had to be increased to 2. Could we add some assertions to ensure they are called with the right params (it looks like one of them should not contain "content_type" nor "topics")?
It would make more explicit what the 2 calls are for

rafahop (commented):
Looks good to me. Just a couple of small comments if you don't mind

rafahop (changes requested):
Please delete accidentally committed `deleteme.txt`

rafahop (approved):

PR #172, opened 2022-10-17
Bump golang.org/text dependency to latest version

nshumoogum:
### What

Dependency upgrade to resolve security vulnerability in package fixed on version 0.38

### How to review

Check ci checks are all passing

### Who can review

Anyone


go.mod
```diff
@@ -4,6 +4,8 @@ go 1.19
 
 replace github.com/spf13/cobra => github.com/spf13/cobra v1.4.0
 
+require github.com/kevinburke/go-bindata v3.24.0+incompatible // indirect
```
rafahop, line 7: Not something you've introduced but should we group this require directive together with the other indirect dependencies?

nshumoogum: So I've updated go.mod so that this require is located next to the rest of the indirect dependencies. I don't know of a command that will actually embed this dependency with the others. Running `go mod tidy` just removes this dependency from go.mod and go.sum.

I'd prefer not to updated go.mod file manually as it will take only one person to run a `go` command to undo a manual change.

rafahop (approved):

PR #173, opened 2022-10-18
update references

NoelleDL:
### What

Update references to `dp-design-system` and `dp-renderer`. 

### How to review

See that reference to `dp-design-system` matches this [hash](https://github.com/ONSdigital/dp-design-system/releases/tag/v2.3.0) and the reference to the `dp-renderer` matches this [release](https://github.com/ONSdigital/dp-renderer/releases/tag/v1.54.0).

### Who can review

Anyone

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -46,7 +46,7 @@ func Get() (*Config, error) {
 	if newCfg.Debug {
 		newCfg.PatternLibraryAssetsPath = "http://localhost:9002/dist/assets"
 	} else {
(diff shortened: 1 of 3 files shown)


nshumoogum (approved):
I think this is okay, my initial concern was inheriting Nicks changes to no index tag in dp-renderer as I thought he had changed the variable name. But it appears he hasn't

This line https://github.com/ONSdigital/dp-renderer/blob/main/assets/templates/main.tmpl#L25; I thought it was changed from

`SearchNoIndexEnabled `

to

`NoIndexEnabled `

But it hasn't; anyway once merged into develop, i will take a look at the html to make sure the search results page still sets the no index tag

PR #174, opened 2022-11-01
Update search topic filters to include census subtopics

nshumoogum:
### What

- Topic cache updated to include subtopic data, labels and release dates as well as ids
- Search results page updated to display the topic filters (english only as this is what is stored in topics API)

### How to review

- Sanity check changes

### Who can review

Anyone

### DIFF
--- a/cache/subtopic.go
+++ b/cache/subtopic.go
@@ -41,8 +41,8 @@ func (t *Subtopics) GetSubtopics(key string) (subtopics []Subtopic) {
 		return
 	}
 
-	t.mutex.RLock()
-	defer t.mutex.RUnlock()
+	t.mutex.Lock()
+	defer t.mutex.Unlock()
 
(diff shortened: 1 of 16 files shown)


cache/subtopic.go
```diff
@@ -5,32 +5,66 @@ import (
...
 	return t.subtopicsMap[key]
 }
 
+// GetSubtopics returns an array of subtopics
+func (t *Subtopics) GetSubtopics(key string) (subtopics []Subtopic) {
+	if t.subtopicsMap == nil {
+		return
+	}
+
+	t.mutex.RLock()
```
RMPall, line 44: (minor): I think you are writing to a map here right ? so this needs to be a ```t.mutex.Lock()``` rather than just a read lock.

RMPall (approved):

PR #179, opened 2023-01-04
Allow search results to be displayed if there is no query param but at least one valid filter set

nshumoogum:
### What

As long as the user has a content type or topic filter set and it is valid then the query parameter can be less than 3 characters (or none at all)

If the filter values are incorrect then return 400 page

### How to review

- Check CI tasks pass
- All changes abide by What above
- No typos etc.

### Who can review

Anyone


NoelleDL (approved):
LGTM

PR #180, opened 2023-01-05
Release - Update Validation of Search Parameters

nshumoogum:
### What

- Allow users to query search with a search term less than 3 characters as long as a filter (content type or topic) is set and is valid
- Improve log message by aligning with rest of logs in app
- Update which version of design system is being used
- Display subtopic ids in url
- Update container base to Ubuntu 20.04

### How to review

- Sanity check changes

### Who can review

Anyone


data/query.go
```diff
@@ -26,33 +28,34 @@ func ReviewQuery(ctx context.Context, cfg *config.Config, urlQuery url.Values, c
...
 
 	reviewSort(ctx, cfg, urlQuery, &validatedQueryParams)
 
-	err = reviewFilters(ctx, urlQuery, &validatedQueryParams)
-	if err != nil {
-		log.Error(ctx, "unable to review filters", err)
-		return validatedQueryParams, err
+	contentTypeFilterError := reviewFilters(ctx, urlQuery, &validatedQueryParams)
+	if contentTypeFilterError != nil {
+		log.Error(ctx, "bad content type filters set", contentTypeFilterError)
```
NoelleDL, line 41: Perhaps a bit nit-picky but maybe "incorrect" or "invalid" would be better here? Up to you.

NoelleDL (approved):

PR #185, opened 2023-01-19
fixed regression where auth no longer works in publishing subnet

allandegnan:
### What

Publishing subnets require auth.  Switching to the SDK broke the auth, this should hopefully restore it.

### How to review

Requires https://github.com/ONSdigital/dp-search-api/pull/184 

### Who can review

!me

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,7 +30,7 @@ require (
 	github.com/ONSdigital/dp-elasticsearch/v3 v3.0.0-alpha.7 // indirect
 	github.com/ONSdigital/dp-mongodb-in-memory v1.3.1 // indirect
 	github.com/ONSdigital/dp-rchttp v1.0.0 // indirect
-	github.com/ONSdigital/dp-search-api v1.29.0 // indirect
+	github.com/ONSdigital/dp-search-api v1.31.0 // indirect
 	github.com/ONSdigital/go-ns v0.0.0-20210916104633-ac1c1c52327e // indirect
 	github.com/aws/aws-sdk-go v1.44.76 // indirect
 	github.com/c2h5oh/datasize v0.0.0-20220606134207-859f65c6625b // indirect
--- a/handlers/handlers.go
+++ b/handlers/handlers.go
@@ -104,10 +104,14 @@ func read(w http.ResponseWriter, req *http.Request, cfg *config.Config, zc Zebed
 			return
 		}
 	}()
-
 	if makeSearchAPICalls {
 		var options searchSDK.Options
+
 		options.Query = searchQuery
+		options.Headers = http.Header{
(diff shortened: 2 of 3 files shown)


handlers/handlers.go
```diff
@@ -183,6 +186,9 @@ func getCategoriesCountQuery(searchQuery url.Values) url.Values {
 func getCategoriesTypesCount(ctx context.Context, accessToken, collectionID string, query url.Values, searchC SearchClient, censusTopicCache *cache.Topic) ([]data.Category, []data.Topic, error) {
 	var options searchSDK.Options
 	options.Query = query
+	options.Headers = map[searchSDK.Header][]string{
+		searchSDK.Authorisation: {"Bearer " + accessToken},
```
nshumoogum, line 190: We need the collection id too.

Do we need to extend the list of available headers in the SDK package to include this? Or can we create our on key value pairs?

nshumoogum (changes requested):

handlers/handlers.go
allandegnan: You don't need to export the actual headers, just the Header struct.  You can write code like this:

``` 
options.Headers = map[searchSDK.Header][]string{
		searchSDK.Authorisation: {"Bearer " + accessToken},
		"header_name": { "header_value" },
```

The list of headers is just a nice to have.

nshumoogum: Yep, I thought as much. So we just need to figure out what the header is that holds the collection id and then add that to the map of headers

allandegnan: Yeah I've updated both PRs now.

https://github.com/ONSdigital/dp-search-api/pull/184

nshumoogum (approved):

go.mod
```diff
@@ -30,7 +30,7 @@ require (
 	github.com/ONSdigital/dp-elasticsearch/v3 v3.0.0-alpha.7 // indirect
 	github.com/ONSdigital/dp-mongodb-in-memory v1.3.1 // indirect
 	github.com/ONSdigital/dp-rchttp v1.0.0 // indirect
-	github.com/ONSdigital/dp-search-api v1.29.0 // indirect
+	github.com/ONSdigital/dp-search-api v1.7.1-0.20230120122810-91b330fb217d // indirect
```
nshumoogum, line 33: We look to only add dependencies of our tagged releases

So will need to get dp-search-api released and tagged and then amend the go.mod file to point to new version here

nshumoogum (changes requested):

go.mod
allandegnan: Fixed now

nshumoogum (approved):

PR #191, opened 2023-02-02
Update census topic count

nshumoogum:
### What

Variable name changed in API response from `distinc_topics_count` to `distinct_items_count`

### How to review

- Check changes make sense

### Who can review

Anyone


data/topic_filter.go
```diff
@@ -15,7 +15,7 @@ import (
 // Topic represents a topic filter on the search page
 type Topic struct {
 	Count              int        `json:"count"`
-	DistinctTopicCount int        `json:"distinct_topics_count"`
```
DavidSubiros, line 18: Maybe we should rename this variable to 'DistinctItemsCount' as well? (or would it cause too many issues?)

DavidSubiros (approved):

PR #195, opened 2023-02-07
Fix handling of census checkbox

nshumoogum:
### What

See title

Previous split on empty string resulted in 1 item being added to the array used for determining if the "census" checkbox should be checked; this resulted in the census checkbox always being checked even if no topic filters were set. See image below for problem this PR fixes

Screenshot 2023-02-07 at 14 02 59

### How to review

Sanity check changes

### Who can review

Anyone

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -5,7 +5,7 @@ run:
   concurrency: 4
 
   # timeout for analysis, e.g. 30s, 5m, default is 1m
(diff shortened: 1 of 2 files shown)


lindenmckenzie (approved):
Code looks fine but lint has failed

lindenmckenzie (approved):

PR #199, opened 2023-02-16
Added handler; Removed the search prefix to reuse partials

allandegnan:
### What

Adds /census/find-a-dataset endpoint.

### How to review

`SEARCH_AB_TEST_PERCENTAGE=100 DATASET_FINDER_ENABLED=1 SEARCH_ROUTES_ENABLED=1 make debug` on dp-frontend-router.

### Who can review

!me


NoelleDL (changes requested):
The design system hash needs to added to`config.go` [](https://github.com/ONSdigital/dp-frontend-search-controller/blob/develop/config/config.go#L49) UPDATE: 24 February - I've updated the search controller with the latest design system so this change isn't needed.

NoelleDL (changes requested):
Just needs to be updated with the latest develop branch as well.

PR #206, opened 2023-03-06
Feature/census find dataset template

allandegnan:
### What

Describe what you have changed and why.

### How to review

Describe the steps required to test the changes.

### Who can review

Describe who worked on the changes, so that other people can review.


handlers/handlers.go
```diff
@@ -31,6 +32,143 @@ func Read(cfg *config.Config, zc ZebedeeClient, rend RenderClient, searchC Searc
...
+	if err != nil {
+		log.Error(ctx, "unable to validate current page", err)
+		setStatusCode(w, req, err)
+		return
+	}
+
+	basePage := rend.NewBasePageModel()
+	m := mapper.CreateDataFinderPage(cfg, req, basePage, validatedQueryParams, categories, topicCategories, populationTypes, dimensions, searchResp, lang, homepageResp, errorMessage, navigationCache)
+	rend.BuildPage(w, m, "search")
+}
```
NoelleDL, line 170: I thought this would render the dataset finder template? I see the `search` templates has been updated but wondering why there isn't a separate dataset finder template.

data/dimensions_filter.go
```diff
@@ -0,0 +1,54 @@
...
+	for i := range dimensions {
+		dimensionsQuery := strings.ToLower(dimensions[i])
+		if dimensionsQuery == "" {
+			continue
+		}
+
+		/*
+		 * TODO - Check population type is valid
+		 */
+
```
NoelleDL, line 34: Think this can be deleted.

mapper/mapper.go
```diff
@@ -51,6 +89,10 @@ func CreateSearchPage(cfg *config.Config, req *http.Request, basePage coreModel.
 
 	mapTopicFilters(cfg, &page, topicCategories, validatedQueryParams)
 
+	mapPopulationTypesFilters(cfg, &page, populationTypes, validatedQueryParams)
+
+	mapDimensionsFilters(cfg, &page, dimensions, validatedQueryParams)
+
 	return page
```
NoelleDL, line 96: Shouldn't there be tests for these?

handlers/handlers.go
allandegnan: The note on the ticket was to save LOC where possible.  None of the features I'm aware of really change the template enough to justify forking it, but if the consensus is it's easier to understand duplicated, I'm happy to do so.

NoelleDL: The ticket says there should be a new template (copied from existing), handler and component tests. Re the refactoring bit to save lines of code... maybe some of the page components could be partials? Linden just did some refactoring in the homepage controller. Can find it if you'd like an example.

config/config.go
```diff
@@ -58,28 +60,30 @@ func get() (*Config, error) {
...
+		DefaultLimit:                            10,
+		DefaultMaximumLimit:                     50,
+		DefaultMaximumSearchResults:             500,
+		DefaultOffset:                           0,
+		DefaultPage:                             1,
+		DefaultSort:                             "relevance",
+		EnableCensusTopicFilterOption:           false,
+		EnableCensusPopulationTypesFilterOption: false,
+		EnableCensusDimensionsFilterOption:      false,
+		EnableNewNavBar:                         false,
```
NoelleDL, line 78: Sorry to pedantic but everything else is in alphabetical order so would be best to follow the convention re `EnableCensusPopulationTypesFilterOption` and `EnableCensusDimensionsFilterOption`

assets/templates/partials/filter.tmpl
```diff
@@ -137,6 +139,108 @@
     </noscript>
 </form>
 
+{{ if $population_types }}
+<form id="filterForm" method="get" class="js-auto-submit__form">
+    <input type="hidden" name="q" value="{{ $query }}"/>
+    <input type="hidden" name="sort" value="{{$sort}}"/>
+    <fieldset class="ons-fieldset">
```
lindenmckenzie, line 146: The separation between each collapsible is quite tight - might be better done as an accordion: 

https://service-manual.ons.gov.uk/design-system/components/accordion#example.accordion.html

assets/locales/service.cy.toml
```diff
@@ -26,6 +26,26 @@ description = "Content type"
...
+
+[Dimensions]
+description = "Dimensions"
+one = "Dimensions"
+other = "Dimensions"
+
+[UsualResidents]
+description = "Usual Residents"
+one = "Usual Residents"
+other = "Usual Residents"
```
lindenmckenzie, line 42: I've put the list of population types on the ticket for you.

assets/templates/partials/filter.tmpl
```diff
@@ -137,6 +139,108 @@
...
+                            name="populations"
+                            value="{{ print $populationFilter.LocaliseKeyName }}" aria-controls="population-group-{{ $index }}-other-wrap"
+                            data-gtm-label="{{ $populationFilter.LocaliseKeyName }}"
+                            {{ if $populationFilter.IsChecked }}
+                                checked
+                            {{ end }}
+                            {{ if eq $populationFilter.Count 0 }} disabled aria-disabled="true" {{ end }}
+                        >
+                        <label class="ons-checkbox__label" for="population-group-{{ $index }}" id="population-group-{{ $index }}-label">
+                            {{ localise $populationFilter.LocaliseKeyName $lang 4 }} ({{ $populationFilter.Count }})
```
lindenmckenzie, line 175: Given we're unlikely to get welsh translations, is it easier to just render whatever the Search API passes?

assets/templates/partials/filter.tmpl
```diff
@@ -137,6 +139,108 @@
     </noscript>
 </form>
 
+{{ if $population_types }}
+<form id="filterForm" method="get" class="js-auto-submit__form">
```
lindenmckenzie, line 143: You've got a few of the same form id here - is there a reason for this? Or should they be part of the same form?

lindenmckenzie: Same question to the hidden query input below.

assets/templates/partials/filter.tmpl
```diff
@@ -137,6 +139,108 @@
...
+                        {{ localise "PopulationTypes" $lang 1 }}
+                    </legend>
+                </h2>
+                <span class="ons-collapsible__icon">
+                    {{ template "icons/chevron-right" }}
+                </span>
+            </summary>
+            <div id="collapsible-content" class="ons-collapsible__content ons-js-collapsible-content">
+                <p class="ons-checkboxes__label">Show only:</p>
+        {{ range $index, $populationFilter := $population_types }}
```
lindenmckenzie, line 160: Think this section needs some indenting.

assets/templates/partials/filter.tmpl
```diff
@@ -137,6 +139,108 @@
...
+                        {{ localise "Dimensions" $lang 1 }}
+                    </legend>
+                </h2>
+                <span class="ons-collapsible__icon">
+                    {{ template "icons/chevron-right" }}
+                </span>
+            </summary>
+            <div id="collapsible-content" class="ons-collapsible__content ons-js-collapsible-content">
+                <p class="ons-checkboxes__label">Show only:</p>
+        {{ range $index, $dimensionFilter := $dimensions }}
```
lindenmckenzie, line 210: Same indenting question as above.

data/dimensions_filter_test.go
```diff
@@ -0,0 +1,46 @@
...
+			{Type: "Ethnicity", Count: 1},
+		},
+		PopulationType: []searchModels.FilterCount{
+			{Type: "Usual Residents", Count: 1},
+		},
+	}
+
+	dimensions := GetDimensions(searchResponseMock)
+
+	Convey("Given the type of the search result", t, func() {
```
lindenmckenzie, line 43: I understand what you're getting at here, but think this Convey statement could be more descriptive.

data/pagination.go
```diff
@@ -192,7 +200,20 @@ func getPageURL(query string, page int, controllerQuery url.Values) (pageURL str
...
+	var u url.URL
+	q := u.Query()
+	q.Set("q", query)
+	if len(populationTypes) > 0 {
+		q.Set("population_types", populationTypes)
+	}
+	if len(dimensions) > 0 {
+		q.Set("dimensions", dimensions)
+	}
+	return q.Encode()
```
lindenmckenzie, line 218: Is this extra separation necessary for populationTypes and dimensions - given that it already happens for contentTypes and topics? Is the purpose of it to ensure the q comes first and the page comes last?

data/population_type_filter_test.go
```diff
@@ -0,0 +1,46 @@
...
+		},
+		PopulationType: []searchModels.FilterCount{
+			{Type: "Usual Residents", Count: 1},
+		},
+	}
+
+	populationTypes := GetPopulationTypes(searchResponseMock)
+
+	Convey("Given the type of the search result", t, func() {
+		So(populationTypes[0].Type, ShouldEqual, searchResponseMock.Items[0].PopulationType)
```
lindenmckenzie, line 44: Same question as on dimensions

data/population_types_filter.go
```diff
@@ -0,0 +1,55 @@
...
+
+type PopulationTypes struct {
+	LocaliseKeyName string `json:"localise_key"`
+	Count           int    `json:"count"`
+	Type            string `json:"string"`
+	Query           string `json:"query"`
+	ShowInWebUI     bool   `json:"show_in_web_ui"`
+}
+
+// reviewTopicFilters retrieves subtopic ids from query, checks if they are one of the census subtopics, and updates validatedQueryParams
```
lindenmckenzie, line 20: This comment needs updating.

data/query_test.go
```diff
@@ -21,12 +21,14 @@ func TestUnitReviewQuerySuccess(t *testing.T) {
...
-			"limit":  []string{"10"},
-			"page":   []string{"1"},
+			"q":                []string{"housing"},
+			"population_types": []string{""},
+			"dimensions":       []string{""},
+			"filter":           []string{"article"},
+			"topics":           []string{"1234,5678"},
+			"sort":             []string{"relevance"},
+			"limit":            []string{"10"},
+			"page":             []string{"1"},
```
lindenmckenzie, line 31: Have you added any query tests for the new params?

lindenmckenzie (commented):
A few comments - happy to discuss

assets/templates/partials/filter.tmpl
allandegnan: I assume the hidden inputs are left over from the days where JS might not have been required, but I checked whats on prod and it requires JS so removing.

data/query_test.go
allandegnan: There's not much to test, but added a check that it's the same after the map function.

data/pagination.go
allandegnan: Sort of.  

This is a hack because we're passing strings of text around rather than the query.

The main aim is to ensure the URL would be encoded with the correct '?' & '&' characters, in the case that anything was omitted.  I'm not sure if the 'q' can / will ever be omit-able.  Probably the "correct" fix is to refactor the entire thing, but it's a bunch of work that's not really part of the ticket.

assets/templates/partials/filter.tmpl
lindenmckenzie: @allandegnan may be better to progress as is and refactor later.

lindenmckenzie (approved):

PR #210, opened 2023-03-08
Feature/remove content filter

allandegnan:
### What

find-a-dataset should only work with 

### How to review

Checkboxes missing, but `filter=dataset_landing_page` sent to api.

### Who can review

!me


lindenmckenzie:
@allandegnan has this PR been superseded or is it still relevant?

allandegnan:
> @allandegnan has this PR been superseded or is it still relevant?

This is the end PR, but needs a fixup and requires this: https://github.com/ONSdigital/dp-search-api/pull/200

config/config.go
```diff
@@ -72,9 +72,9 @@ func get() (*Config, error) {
 		DefaultOffset:                           0,
 		DefaultPage:                             1,
 		DefaultSort:                             "relevance",
-		EnableCensusTopicFilterOption:           false,
-		EnableCensusPopulationTypesFilterOption: false,
-		EnableCensusDimensionsFilterOption:      false,
+		EnableCensusTopicFilterOption:           true,
+		EnableCensusPopulationTypesFilterOption: true,
+		EnableCensusDimensionsFilterOption:      true,
```
lindenmckenzie, line 77: Don't think these should be set to true by default yet.

lindenmckenzie (commented):
Other than the config bools, this looks fine.

config/config.go
nshumoogum: I was about to make the same comment, these should be false and then environment secrets containing these flags should be updated

features/dataset_finder.feature
lindenmckenzie, line 1: Accepted deletion of test as this feature is not here to retro-fit mocks to everything in search controller. Task to be picked up as tech debt.

lindenmckenzie (approved):

PR #218, opened 2023-03-24
allow screenreader to notify user of an expandable group

NoelleDL:
### What

Fix for AA accessibility violations picked up by axe. 
- [ticket 1261](https://trello.com/c/Rn2PAdn8/1261-incorrect-aria-attribute)

### How to review

See that tests pass and code makes sense

### Who can review

FE dev


config/config.go
```diff
@@ -72,7 +72,7 @@ func get() (*Config, error) {
 		DefaultOffset:                           0,
 		DefaultPage:                             1,
 		DefaultSort:                             "relevance",
-		EnableCensusTopicFilterOption:           false,
+		EnableCensusTopicFilterOption:           true,
```
nshumoogum, line 75: This should be set to false and we enable this flag in each environment

nshumoogum (changes requested):

assets/templates/partials/filter.tmpl
```diff
@@ -13,10 +13,10 @@
         <input type="hidden" name="sort" value="{{$sort}}"/>
         <fieldset class="ons-fieldset">
             {{ range $index, $topicFilter := $topicFilters }}
-                <div class="ons-checkboxes__items">
+                <div class="ons-checkboxes__items" role="tree">
                     <span class="ons-checkboxes__item ons-checkboxes__item--no-border">
                         <span class="ons-checkbox ons-checkbox--no-border topic-filter">
-                            <input type="checkbox" id="topic-group-{{ $index }}"
+                            <input type="checkbox" role="group" id="topic-group-{{ $index }}"
```
lindenmckenzie, line 19: I'm not sure this group role is allowed on an input - I can't get at an Axe extension but I've run it through an axe crawler and w3c html checker and they both complain about it.

NoelleDL: Hmmm. Mine didn't complain when I ran. Will look again.

PR #231, opened 2023-03-29
dataset: set a default sort to date, removed relevance

allandegnan:
### What

Changed the data to sort by date by default

### How to review

Enable flag, look at the page, check sort is date, then select a drop down to ensure it changes.

### Who can review

!me


lindenmckenzie (changes requested):
Unit test failure - otherwise ok.

PR #232, opened 2023-03-30
Fix welsh pluralisations

lindenmckenzie:
### What

Fixes issue with pluralisations that affects welsh not having translations for 'few', 'many' etc caused by using number of results as a count mechanism. Not sure why this only affects welsh. 

### How to review

Load frontend-search-controller and connect to sandbox search-api. 

Need to run `document.cookie="lang=cy"` in the console and refresh to have welsh. 

### Who can review

Not me.


assets/templates/partials/count.tmpl
```diff
@@ -3,5 +3,9 @@
 <h2>
     {{ $numberResults := len .Data.Response.Items}}
     <span class="search__summary__count">{{ .Data.Response.Count}} </span>
-    {{ localise .Data.TermLocalKey $lang $numberResults }}
+    {{ if eq $numberResults 1 }}
+        {{ localise .Data.TermLocalKey $lang 1 }}
+    {{ else }}
+        {{ localise .Data.TermLocalKey $lang 4 }}
+    {{ end }}
 </h2>
```
mr-nick17, line 11: Empty EOF line is missing

assets/templates/partials/count.tmpl
```diff
@@ -3,5 +3,9 @@
 <h2>
     {{ $numberResults := len .Data.Response.Items}}
     <span class="search__summary__count">{{ .Data.Response.Count}} </span>
-    {{ localise .Data.TermLocalKey $lang $numberResults }}
```
mr-nick17, line 6: This probably isn't working because the welsh translations are missing from the other counts (likely 2 and 3; but I can't tell from the PR what the key is to look)

mr-nick17 (approved):

PR #243, opened 2023-04-27
Fix publishing preview

jondewijones:
### What

Fix publishing preview
Update to use v2.0.3 of `dp-renderer` (contains new publishing preview template)
Map `IS_PUBLISHING` feature flag to template data

⚠️⚠️ not been tested locally!! Getting search running locally with `IS_PUBLISHING=true` proved very tricky - lots of cached navigation errors

### Who can review

Anyone


go.mod
```diff
@@ -5,22 +5,21 @@ go 1.19
...
 	github.com/ONSdigital/dp-cache v0.1.0
 	github.com/ONSdigital/dp-component-test v0.7.0
 	github.com/ONSdigital/dp-cookies v0.3.3
-	github.com/ONSdigital/dp-healthcheck v1.5.0
-	github.com/ONSdigital/dp-net v1.5.0
-	github.com/ONSdigital/dp-net/v2 v2.8.1
-	github.com/ONSdigital/dp-renderer v1.62.0
+	github.com/ONSdigital/dp-healthcheck v1.6.1
+	github.com/ONSdigital/dp-net/v2 v2.9.1
+	github.com/ONSdigital/dp-renderer/v2 v2.0.3
```
lindenmckenzie, line 14: One of the breaking changes from v2 affects the breadcrumbs used in the search.tmpl file. 

https://github.com/ONSdigital/dp-renderer/releases/tag/v2.0.1

lindenmckenzie: Has lead to double breadcrumb (can be seen by running in Web mode and accessing /census/find-a-dataset

lindenmckenzie (changes requested):
Upgrade to v2 has led to a double breadcrumb on dataset finder - other changes look sound, spent some time trying to make it work locally to test but was unable to. Make sure is tested in sandbox - we might want to do further digging under another ticket as to why it's not working locally.

go.mod
jondewijones: Good spot. Remove breadcrumbs partial

PR #252, opened 2023-07-26
Feature/switch to user auth

lindenmckenzie:
### What

Switched to user auth for publishing mode (as is standard)

### How to review

Can run search publishing stack in dp-compose, review code, check tests pass. 

### Who can review

Not me.


jondewijones (approved):
LGTM

PR #263, opened 2023-11-13
add condition for token prefix

andre-urbani:
### What

Add condition for adding token prefix
Fixes the issue described in this ticket - https://trello.com/c/cckHOSbg/1844-ensure-search-controller-works-in-publishing-mode-with-sec-auth-on-and-off

### How to review

Confirm changes make sense, tests pass

### Who can review

Anyone


jondewijones (commented):
Since we repeat the `if strings.HasPrefix(accessToken...` logic three times do you think it's worth moving to a function to reduce duplication?

handlers/handlers.go
```diff
@@ -425,3 +432,11 @@ func setStatusCode(w http.ResponseWriter, req *http.Request, err error) {
 
 	w.WriteHeader(status)
 }
+
+func setFlorenceTokenHeader(headers http.Header, accessToken string) {
```
jondewijones, line 436: Is it worth writing a test for this?

jondewijones (approved):

PR #271, opened 2023-12-20
Fix/mobile view aggregation

lindenmckenzie:
### What

Removed the archive message accordion from mobile view
Added show results button to mobile to be able to close the modal
Added the updated and keyword inputs to the mobile filter modal
Updated formatting on date input

### How to review

Check looks ok and behaves as expected - will need to port forward to dp-api-router.

### Who can review

Frontend Go Dev.


mr-nick17 (approved):
The date input fields could come from `dp-renderer` but appreciate that updating models, etc. would be quite cumbersome

PR #279, opened 2024-01-23
Add AB test handlers for nlp weighting

lindenmckenzie:
### What

Added AB testing functionality for NLP weighting. 

### How to review

Check tests pass, code looks good. 

Optionally:
- port forward to api router
- set ab_test cookie to appropriate value
- see the requests coming in to search-api via kibana with nlp_weighting=true vs false
- check the gtm data layer script to see different values being assigned to "abTest"

### Who can review

Frontend Go dev.


handlers/handlers.go
```diff
@@ -355,6 +362,12 @@ func read(w http.ResponseWriter, req *http.Request, cfg *config.Config, zc Zebed
 		return
 	}
 
+	validatedQueryParams.NLPWeightingEnabled = nlpWeightingEnabled
+
```
mr-nick17, line 366: The log is related to the function so I think there should be no whitespace

handlers/handlers_test.go
```diff
@@ -180,14 +180,30 @@ func TestUnitReadSuccess(t *testing.T) {
...
+			})
+		})
+
+		Convey("When read is called with NLP switched on", func() {
+			read(w, req, cfg, mockedZebedeeClient, mockedRendererClient, mockedSearchClient, accessToken, collectionID, englishLang, *mockCacheList, "search", true)
+
+			Convey("Then a 200 OK status should be returned", func() {
+				So(w.Code, ShouldEqual, http.StatusOK)
+
+				So(len(mockedRendererClient.BuildPageCalls()), ShouldEqual, 1)
```
mr-nick17, line 202:
```suggestion
				So(mockedRendererClient.BuildPageCalls(), ShouldHaveLength, 1)
```

handlers/handlers_test.go
```diff
@@ -180,14 +180,30 @@ func TestUnitReadSuccess(t *testing.T) {
...
+				So(len(mockedSearchClient.GetSearchCalls()), ShouldEqual, 2)
```
mr-nick17, line 203: ..and here

handlers/handlers_test.go
```diff
@@ -180,14 +180,30 @@ func TestUnitReadSuccess(t *testing.T) {
...
+				So(len(mockedZebedeeClient.GetHomepageContentCalls()), ShouldEqual, 1)
```
mr-nick17, line 204: and here

handlers/handlers_test.go
```diff
@@ -239,7 +255,7 @@ func TestUnitReadFailure(t *testing.T) {
 		So(err, ShouldBeNil)
 
 		Convey("When read is called", func() {
-			read(w, req, cfg, mockedZebedeeClient, mockedRendererClient, mockedSearchClient, accessToken, collectionID, englishLang, *mockCacheList, "search")
+			read(w, req, cfg, mockedZebedeeClient, mockedRendererClient, mockedSearchClient, accessToken, collectionID, englishLang, *mockCacheList, "search", false)
```
mr-nick17, line 258: Not for here (maybe) but I think these function signatures for all the handlers are getting quite big and are all quite similar. Do you think they should be a method with a receiver type?

mr-nick17 (commented):
Some minor comments/suggestions

handlers/handlers_test.go
lindenmckenzie: Take the point but later :)

mr-nick17 (approved):

PR #291, opened 2024-03-25
add aria-live attributes to search and data agrregation results

jondewijones:
### What

add aria-live attributes to search and data agrregation results

### Who can review

Anyone


lindenmckenzie (changes requested):
You've missed /timeseriestool here I think?

lindenmckenzie (approved):

PR #296, opened 2024-04-04
Feature/content type filters redesign

StRhys:
### What

This [ticket](https://trello.com/c/sk5rB5S1) was to bring the content type filters inline with the designs.

Checking and unchecking the tickboxes produces some weird behaviour though, such as

- On `/publications` checking all filters, starting with 'Statistical bulletins', will only filter to 'bulletins` and have 4514 rather than the expected 7475
- Checking 'Compendiums' doesn't do anything, in the chrome network tab, it's causing a 500 error
- Going to http://localhost:25000/publications?page=1&filter=article&filter=compendia is ok (minus filtering not working) but going to http://localhost:25000/publications?page=1&filter=compendia throws the 500 page

Looks like the api is only looking at the first filter query and ignoring the others but then works fine on `/search` 
This issue might be slightly out the scope of this ticket, but happy to be pointed in the correct direction to solve it, or a ticket thats already in the backlog.

### How to review

Pull this branch and run app make debug ENABLE_REWORKED_DATA_AGGREGATION_PAGES=true
Run dp-design-system
Port forward the api router to sandbox dp ssh sandbox web 1 -p 23200:localhost:10800
Pages:

https://dp.aws.onsdigital.uk/datalist
https://dp.aws.onsdigital.uk/publications

Toggle content type filters, refresh
Sense check.

### Who can review

Frontend


assets/templates/partials/data-filters/single-content-type-filter.tmpl
```diff
@@ -0,0 +1,50 @@
...
+                    {{ $id := index $childFilter.FilterKey 0 }}
+                    <div class="ons-checkboxes__items">
+                        <span class="ons-checkboxes__item ons-checkboxes__item--no-border">
+                            <span class="ons-checkbox ons-checkbox--no-border child-filter">
+                                <input id="{{ $id }}"
+                                    class="ons-checkbox__input ons-js-checkbox"
+                                    type="checkbox" name="filter"
+                                    {{ if $childFilter.IsChecked }}checked{{ end }}
+                                    data-gtm-label="{{ $childFilter.LocaliseKeyName }}"
+                                    value="{{ $id }}"
```
lindenmckenzie, line 36: this is the one that's tripping up the compendia I'd guess.

lindenmckenzie (changes requested):
Have not reviewed the code. 

Item one:

> On /publications checking all filters, starting with 'Statistical bulletins', will only filter to 'bulletins` and have 4514 rather than the expected 7475

On this, it's because the `dp-search-api` is expecting the `content_type` param to be one param not multiples. This is not a change you've introduced :) 

Have a look into it and see whether you think we should add it to the scope of this ticket or add a new one. Whether we correct it here or in the api is something you should think about too. 

> Checking 'Compendiums' doesn't do anything, in the chrome network tab, it's causing a 500 error

Log from the search API is "content type(s) not allowed" - think it should be a `compendium_landing_page`

This is down to where you're setting the value in the input field. Because of item 1, it gets ignored when it's not on it's own.

StRhys:
After lots of trials and missteps, I found the issue, a missing function which changed the filters into the corresponding content_types. Was on the search query but not the aggregation query. So this sorts both issues!

One issue I have spotted though is on `/datalist`, the `datasets` filter shows two different results, 4287 for the filter, but 4282 for the total. See image below. The `datasets` filter only has two content_types, being `dataset_landing_page` and `timeseries_dataset` and looking at the api, these only return 4282. From what I can see when getting the filters/counts it's only fetching the results for `dataset_landing_page` and `timeseries_dataset`, which is returning the 4287, so unsure where these extra 5 for the filter are coming from.

image

mr-nick17 (commented):
Tested and looks good 👍. Minor markup comment

mr-nick17 (approved):

lindenmckenzie (approved):

PR #316, opened 2024-05-22
Fix errors on high page count

lindenmckenzie:
### What

Fixed 500 / 400 errors on bad params. Instead map them to the page to allow rendering of errors. Also added tests for data aggregation pages handlers. 

### How to review

Check looks ok, tests pass. Optionally:
- switch on aggregated data pages
- port forward to sandbox api router
- access /timeseriestool or /alladhocs and try a page limit higher than 50 (51 / 52 in prod throw 400s, 53+ throws 500)

### Who can review

not me.

### DIFF
--- a/handlers/handlers.go
+++ b/handlers/handlers.go
@@ -18,6 +18,7 @@ import (
 	"github.com/ONSdigital/dp-frontend-search-controller/config"
 	"github.com/ONSdigital/dp-frontend-search-controller/data"
 	"github.com/ONSdigital/dp-frontend-search-controller/mapper"
+	"github.com/ONSdigital/dp-frontend-search-controller/model"
 	dphandlers "github.com/ONSdigital/dp-net/v2/handlers"
 	searchModels "github.com/ONSdigital/dp-search-api/models"
 	searchSDK "github.com/ONSdigital/dp-search-api/sdk"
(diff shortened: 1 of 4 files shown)


handlers/handlers.go
```diff
@@ -261,12 +262,10 @@ func readDataAggregation(w http.ResponseWriter, req *http.Request, cfg *config.C
...
-		if vErr.ID != DateFromErr && vErr.ID != DateToErr {
-			log.Error(ctx, "unable to review query", errors.New(vErr.Description.Text))
-			setStatusCode(w, req, errors.New(vErr.Description.Text))
-			return
-		}
+	if len(validationErrs) > 0 {
+		m := mapper.CreateDataAggregationPage(cfg, req, rend.NewBasePageModel(), validatedQueryParams, []data.Category{}, []data.Topic{}, &searchModels.SearchResponse{}, lang, zebedeeCli.HomepageContent{}, "", navigationCache, template, topicModels.Topic{}, validationErrs)
+		buildDataAggregationPage(w, m, rend, template)
+		return
 	}
```
Oluwawunmi, line 269: - I think a little comment on what the review does what you're doing with the errors would be helpful.
- Logging the errors at some point might also not be a bad idea.

Oluwawunmi (commented):
Just a small feedback

Oluwawunmi (approved):
Lovely.

PR #325, opened 2024-06-04
Release/0.57.1

StRhys:
### What

- [Fix](https://github.com/ONSdigital/dp-frontend-search-controller/pull/318) rss feed link

### Who can review

Anyone


mapper/mapper.go
```diff
@@ -103,12 +111,12 @@ func mapDataPage(page *model.SearchPage, respC *searchModels.SearchResponse, lan
...
 		page.Data.SingleContentTypeFilterEnabled = true
 		page.Data.DateFilterEnabled = true
-		page.RSSLink = fmt.Sprintf("datalist?rss&%s", req.URL.RawQuery)
+		setRSSLink(page, "datalist", req.URL.RawQuery)
 	case "home-publications":
 		page.Metadata.Title = "Publications"
 		page.Title.LocaliseKeyName = "HomePublications"
 		page.Data.SingleContentTypeFilterEnabled = true
-		page.RSSLink = fmt.Sprintf("publications?rss&%s", req.URL.RawQuery)
+		setRSSLink(page, "publications", req.URL.RawQuery)
```
Oluwawunmi, line 119: Is there a reason why the setRSSLink is only set on just datalist and publications? I believe the rss works for all the aggregated pages and not just those two.

Oluwawunmi (commented):
Just a little clarification.

mapper/mapper.go
StRhys: I believe it was a carry over from the old sites, only `/datalist` and `/publications`, so only those two [needed](https://trello.com/c/L8pZ7NAU) it.

Oluwawunmi: [Staticlist](https://www.ons.gov.uk/staticlist?rss)
[Adhocs](https://www.ons.gov.uk/alladhocs?rss)

With the impl., 8 of them work with the rss feature, so what do we do?

Oluwawunmi (approved):

PR #327, opened 2024-06-10
Filter unrecognised filter logging for known unused filters

lindenmckenzie:
### What

Added a known list of unused content type filters - this will reduce overall logging by the app by around 300K log entries a day. 

### How to review

Run the app, port forward to api router in aws and check for 'unrecognised filter type returned from api'. Should be none. Check remaining filters still work. 

### Who can review

Not me.

### DIFF
--- a/data/filter.go
+++ b/data/filter.go
@@ -70,6 +70,10 @@ var (
 		ContentTypes:    []ContentType{Methodology, CorporateInformation, ProductPage},
 	}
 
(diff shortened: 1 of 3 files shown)


AWitcherONS (approved):
No feature Test?

lindenmckenzie:
All it changes is the logging so I didn't feel that it needed a feature level test.

PR #329, opened 2024-06-11
Fix/remove excess logging2

lindenmckenzie:
### What

Removed more excess logging statements - will remove around 420K logs per day. 

### How to review

Port forward to web api router, make a generic request (/search?q=test) and ensure these logs don't appear. Then make requests with strange options (limit=pear, page=grapefruit, sort=rollercoasters) and ensure they do appear.  

### Who can review

Not me.


gedge (approved):
good stuff

PR #347, opened 2024-07-09
Refactor RSS link creation

lindenmckenzie:
### What

Refactored RSS link generation - makes it simpler, and also I'm suspicious of this function in the mem leak investigation.

### How to review

Check generates RSS link correctly - switch on agg pages and go to:
/publications
/data

### Who can review

Frontend go dev.


StRhys (approved):
Approved but minor issue for later I just spotted, unlike the search input, when adding a content type, the rss feed link doesn't update until refresh. Then again could it be that functionality adding to memory leak.

PR #348, opened 2024-07-10
Upgrade ds - guard for no RSS link

lindenmckenzie:
### What

Upgrade DS to add guard for no RSS link on timeseriestool

### How to review

Check ok - works with DS. Commits match. 

### Who can review

Not me.


config/config.go
```diff
@@ -100,8 +100,8 @@ func get() (*Config, error) {
 		EnableCensusTopicFilterOption:           false,
 		EnableCensusPopulationTypesFilterOption: false,
 		EnableCensusDimensionsFilterOption:      false,
-		EnableAggregationPages:                  false,
-		EnableTopicAggregationPages:             false,
+		EnableAggregationPages:                  true,
```
Oluwawunmi, line 103: I believe these two needs to be set back to false

Oluwawunmi (changes requested):

config/config.go
lindenmckenzie: rugh!

lindenmckenzie: teach me for using `-am`

Oluwawunmi (approved):

PR #355, opened 2024-09-03
Fix aggregated pages in publishing mode

lindenmckenzie:
### What

Fix agg pages in publishing mode when logged in to florence.

### How to review

Check consistency with approach for main search (which works). 

To test:

Populate local search index, run florence in SecAuth mode (`SharedConfig.EnableNewSignIn`) and connect to sandbox api router for auth / search api / zebedee. 

Login to florence (to get auth cookies) and then access aggregated page routes (/alladhocs, /economy/publications). Requires `EnableTopicAggregationPages` and `EnableAggregationPages` to be switched on. 

### Who can review

Not me.


Oluwawunmi (approved):
LGTM

PR #357, opened 2024-09-04
MVP page for /previousreleases

jondewijones:
### What

**Note:** this is an MVP and lots has been copy and pasted from other parts. This will get tidied up and/or bits removed and/or bits added once we have a finished design for the page and know what the pages needs/doesn't need

**Note:** the page currently displays all search results until search API can return previous releases

- Add /previousrelease handling (handler, mapper, template, etc)
- Update api-clients library
- Update [build tags](https://stackoverflow.com/questions/68360688/whats-the-difference-between-gobuild-and-build-directives) in makefile. Was causing errors when trying to build assets during testing

image

### How to review

Sanity check, test rendering, etc.

### Who can review

Anyone


cookel2 (approved):
I added a couple of small comments but I'm approving as I've tested it successfully, thanks.

PR #372, opened 2024-10-08
Fix submit not working as expected

cookel2:
### What

Link to Jira ticket: https://jira.ons.gov.uk/browse/DIS-1568

To see the current problems first go to the 'Published data' list here: https://www.ons.gov.uk/datalist

Problem 1:
Entering a word (e.g. "cheese") into the 'Search keywords' text box, on the 'Published data' list, currently gives a URL containing many parameters and two of them are named "q":

https://www.ons.gov.uk/datalist?q=cheese&after-day=&after-month=&after-year=&before-day=&before-month=&before-year=&q=&sort=release_date

This change removes the "q" parameter from the right hand side, as it is not being used for anything, and renames the "q" on the left to be called "kw-query". This is because there are already several other "q" parameters in use on the same page.

Problem 2: 
Ticking 2 of the filter check boxes on the left (e.g. 'Time series' and 'Datasets') then clicking the green button gives you an initial URL like this, which is correct so far:

https://www.ons.gov.uk/datalist?q=&filter=time_series&filter=datasets&after-day=&after-month=&after-year=&before-day=&before-month=&before-year=&q=&sort=release_date

If you then untick the 'Datasets' filter (for example) and click the green button again then its check box is automatically re-ticked and both filters are incorrectly re-added to the right hand side of the URL, like this:

ttps://www.ons.gov.uk/datalist?q=&filter=time_series&after-day=&after-month=&after-year=&before-day=&before-month=&before-year=&q=&filter=datasets&filter=time_series&sort=release_date

### How to review

In your local environment do as follows:

- Run the dp-frontend-search-controller:
`make debug API_ROUTER_URL=https://api.dp.aws.onsdigital.uk/v1 ENABLE_AGGREGATION_PAGES=true`
- Run the dp-frontend-router:
`make debug API_ROUTER_URL=https://api.dp.aws.onsdigital.uk/v1 SEARCH_ROUTES_ENABLED=true DATA_AGGREGATION_PAGES_ENABLED=true`
- Run the dp-design-system:
`make debug`
- You should then see the sandbox list of published data here:
http://localhost:20000/datalist
- Enter a word to search for in the 'Search keywords' textbox e.g. Cheese
- Click the green button next to the textbox
- Observe that the URL should now look something like this (problem 1 is fixed):

http://localhost:20000/datalist?kw-query=cheese&after-day=&after-month=&after-year=&before-day=&before-month=&before-year=&sort=release_date

- Tick the 'Time series' and 'Datasets' check boxes, click the green button again, to change the URL to this:

http://localhost:20000/datalist?kw-query=cheese&filter=time_series&filter=datasets&after-day=&after-month=&after-year=&before-day=&before-month=&before-year=&sort=release_date

- Untick 'Datasets' and click the green button again. Observe that 'Datasets' remains unticked and is removed from the URL (problem 2 is fixed):

http://localhost:20000/datalist?kw-query=cheese&filter=time_series&after-day=&after-month=&after-year=&before-day=&before-month=&before-year=&sort=release_date

### Who can review

Anyone.


mapper/mapper.go
```diff
@@ -182,7 +182,7 @@ func mapDataPage(page *model.SearchPage, respC *searchModels.SearchResponse, lan
 
 	page.Data.KeywordFilter = coreModel.CompactSearch{
 		ElementId: "keywords",
-		InputName: "q",
+		InputName: "kw-query",
 		Language:  lang,
 		Label: coreModel.Localisation{
 			LocaleKey: "SearchKeywords",
```
lindenmckenzie, line 188: I don't think this is the right way to go about this - the javascript is for example still updating the q parameter. 

Having tested it, I also don't think it's necessary. This fix was the one in the .tmpl file I think as there's an extra input there which would've been confusing things.

lindenmckenzie (changes requested):

mapper/mapper.go
cookel2: Yes I changed this to kw-query more as a 'nice to have' as there are 4 other parameters all named "q" on the same page (one for main search in the title bar, one for filters, one for sorts, and one for pagination. Having changed this it was easier to discover the fix in the tmpl file because I could see that it was one of the parts that was still using "q". I can change it back to "q" if you prefer though?

cookel2: 👍 ok that makes sense, thanks

lindenmckenzie (approved):

PR #373, opened 2024-10-09
Add parent content type setter func

jondewijones:
### What

Set the parent page data type on the API query to fix return compendia and compendia chapters in `/previousreleases` page

image

### Who can review

Anyone

### DIFF
--- a/data/query.go
+++ b/data/query.go
@@ -11,6 +11,7 @@ import (
 	"github.com/ONSdigital/dp-frontend-search-controller/cache"
 	"github.com/ONSdigital/dp-frontend-search-controller/config"
 	"github.com/ONSdigital/log.go/v2/log"
+	"github.com/davecgh/go-spew/spew"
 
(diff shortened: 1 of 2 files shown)


data/query.go
```diff
@@ -163,7 +164,7 @@ func ReviewPreviousReleasesQueryWithParams(ctx context.Context, cfg *config.Conf
 	paginationErr := reviewPagination(ctx, cfg, urlQuery, &sp)
 	validationErrs = handleValidationError(ctx, paginationErr, "unable to review pagination for previous releases", PaginationErr, validationErrs)
 	reviewSort(ctx, urlQuery, &sp, cfg.DefaultPreviousReleasesSort)
-
+	spew.Dump(sp)
```
mr-nick17, line 167: can we remove the log dump?

jondewijones: 🤦. Good spot. Removed

mr-nick17 (approved):

PR #387, opened 2024-10-25
Fix/search sort dropdown

StRhys:
### What

This was to fix the search dropdown when js is disabled.

I also fixed an issue with search query not being retained when js is disabled and the apply filter button is pressed.

I've made the search page more resemble the aggregated pages.
- Apply filter button at bottom of filters
- Removed individual buttons from filters when js is disabled
- Wrapped all filters in same form tags
- Removed form tags from each filter

### How to review

Pull this branch
Run dp-frontend-search-controller using make debug ENABLE_REWORKED_DATA_AGGREGATION_PAGES=true
Port forward the api router to sandbox with `dp ssh sandbox web 1 -p 23200:localhost:10800`

- http://localhost:25000/search?q=climate

check all filters and sort work.

Turn js off and refresh, and check they still work.

### Who can review

Frontend

### DIFF
--- a/assets/templates/data-aggregation-page.tmpl
+++ b/assets/templates/data-aggregation-page.tmpl
@@ -59,7 +59,7 @@
           <div class="search__count">
             {{ template "partials/count" . }}
           </div>
-          <div class="ons-grid ons-u-mt-s">
+          <div class="ons-grid ons-u-mt-s ons-u-mb-s">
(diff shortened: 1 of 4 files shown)


mr-nick17:
Global search and sort is working without js 🎉 
But the margin has also gone from the data aggregation pages 🤦‍♂️ 
margin removed from sort

mr-nick17 (approved):

PR #388, opened 2024-10-31
Fix/invalid params cause oom

mr-nick17:
### What

✅ [Fix invalid page parameters causing app OOM](https://jira.ons.gov.uk/browse/DIS-1999)
- Refactor ordering of api calls and where concurrent calls happen
- Added missing page mapping for validation errors
- Added missing toggle from template to handle missing search data

### How to review

Sense check
CI checks pass

image

Check locally
- Port forward to sandbox api router `dp ssh sandbox web 1 -p 23200:localhost:10800`
- Run `dp-design-system` `nvm use && npm run dev`
- Pull this branch and run controller `make debug`
- See the error for yourself at: http://localhost:25000/employmentandlabourmarket/peopleinwork/employmentandemployeetypes/bulletins/uklabourmarket/previousreleases?page=1000 

### Who can review

!me


assets/templates/related-list-pages.tmpl
```diff
@@ -1,27 +1,31 @@
...
+          {{ template "partials/error-summary" .Error }}
+        </div>
+      {{ else }}
+        <section
+          class="search__summary"
+          role="contentinfo"
+        >
+          {{ template "partials/type-label" . }}
+          {{ template "partials/related-list-pages-title" . }}
+          <p>{{ .Metadata.Description }}</p>
```
lindenmckenzie, line 15: we're removing the whole page if there's errors?

handlers/handlers.go
```diff
@@ -395,121 +395,134 @@ func readPreviousReleases(w http.ResponseWriter, req *http.Request, cfg *config.
...
-		homepageResp, homeErr = zc.GetHomepageContent(ctx, accessToken, collectionID, lang, homepagePath)
-		if homeErr != nil {
-			log.Warn(ctx, "unable to get homepage content", log.FormatErrors([]error{homeErr}))
+		// check page type
+		pageData, pdErr = zc.GetPageData(ctx, accessToken, collectionID, lang, latestContentURL)
+		if pdErr != nil {
+			log.Error(ctx, "failed to get content type", pdErr)
+			cancel()
+			return
+		}
```
lindenmckenzie, line 426: The reason this was structured non-concurrently was to prevent additional calls if we shouldn't render the page at all.

handlers/handlers.go
```diff
@@ -395,121 +395,134 @@ func readPreviousReleases(w http.ResponseWriter, req *http.Request, cfg *config.
...
+	if len(validationErrs) > 0 {
+		log.Info(ctx, "validation of parameters failed for aggregation", log.Data{
+			"parameters": validationErrs,
+		})
+		// Errors are now mapped to the page model to output feedback to the user rather than
+		// a blank 400 error response.
+		m := mapper.CreatePreviousReleasesPage(cfg, req, rend.NewBasePageModel(), validatedQueryParams, &searchModels.SearchResponse{}, lang, homepageResp, "", navigationCache, template, cache.Topic{}, validationErrs, pageData, bc)
+		rend.BuildPage(w, m, template)
+		return
+	}
```
lindenmckenzie, line 491: I'm not sure about the ordering on this, let's discuss.

handlers/handlers.go
```diff
@@ -395,121 +395,134 @@ func readPreviousReleases(w http.ResponseWriter, req *http.Request, cfg *config.
...
+	options.Query = searchQuery
+
+	options.Headers = http.Header{
+		searchSDK.FlorenceToken: {Bearer + accessToken},
+		searchSDK.CollectionID:  {collectionID},
+	}
+
+	searchResp, respErr := searchC.GetSearch(ctx, options)
+	if respErr != nil {
+		log.Error(ctx, "getting search response from client failed for aggregation", respErr)
```
lindenmckenzie, line 510: Not aggregation page.

lindenmckenzie (commented):
Think we should discuss some of the logic ordering here.

handlers/handlers.go
mr-nick17: We were never handling the error, just logging it. This was the page still renders

handlers/handlers.go
mr-nick17: 🤦‍♂️ - foiled by own copying and pasting

assets/templates/related-list-pages.tmpl
mr-nick17: yep the partials require some mapped data (search response I believe) that is causing the app to OOM as the mapping fails (especially for the higher page params i.e 100) and the app doesn't know how to handle and by design we just let the app OOM 😄

mr-nick17:
As discussed. Will break out the `GetPageData` call from concurrency and will display page title and summary for page with validation error

handlers/handlers.go
```diff
@@ -395,121 +395,123 @@ func readPreviousReleases(w http.ResponseWriter, req *http.Request, cfg *config.
...
 	}
 
-	basePage := rend.NewBasePageModel()
-	err = validateCurrentPage(ctx, cfg, validatedQueryParams, searchResp.Count)
-	if err != nil {
+	sanitisedParams := sanitiseQueryParams(allowedPreviousReleasesQueryParams, urlQuery)
+
+	validatedQueryParams, validationErrs := data.ReviewPreviousReleasesQueryWithParams(ctx, cfg, sanitisedParams, urlPath)
+	if len(validationErrs) > 0 {
+		log.Info(ctx, "validation of parameters failed for aggregation", log.Data{
```
lindenmckenzie, line 472:
```suggestion
		log.Info(ctx, "validation of parameters failed for previous releases", log.Data{
```

lindenmckenzie (changes requested):

lindenmckenzie (approved):

PR #396, opened 2024-11-14
fix/convert div to p tag

mr-nick17:
### What

Convert `div` to `p` tag

### How to review

Sense check

### Who can review

!me


dhee-tree (approved):
LGTM

PR #400, opened 2024-11-21
Add datasetID if present

lindenmckenzie:
### What

Added Dataset ID if present to results list pages.

I haven't extended tests with these as we currently don't check what they look like anyway. We should do this better in future but I don't think this is the PR to do it in. 

### How to review

Run service, port forward to sandbox api router and run a related data page with datasets on - see visible dataset IDs and it not appearing if they're not there. 

### Who can review

Describe who worked on the changes, so that other people can review.

### DIFF
--- a/assets/templates/partials/list.tmpl
+++ b/assets/templates/partials/list.tmpl
@@ -47,7 +47,12 @@
                                 {{ end }}
                             {{ else }}
                                 <li class="ons-document-list__item-attribute">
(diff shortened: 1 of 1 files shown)


cookel2 (approved):
I've tested it locally and the dataset IDs are appearing if present. However, I notice that the word 'dataset' looks a little out of place. It appears between the information about the release date and the ID and looks like it should be on the left hand side with a colon after it (or something like that).

PR #425, opened 2025-05-14
Downgrade topic path error to info

lindenmckenzie:
### What

Downgraded invalid topic path to info from error as it is a frequent user behaviour due to how we manage the routing. 

### How to review

Check looks ok. 

### Who can review

Not me.


redhug1 (approved):
LGTM

